PR #8, opened 2024-01-31
Add metric types

flora-hofmann-frequenz:
This PR adds `_types.py` for all `.proto` definitions that have been used in the reporting-api.

### DIFF
--- a/src/frequenz/client/common/metrics/__init__.py
+++ b/src/frequenz/client/common/metrics/__init__.py
@@ -1,0 +1,407 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Module to define the metrics used with the common client."""
+
+from __future__ import annotations  # required for constructor type hinting
+
+import enum
+from dataclasses import dataclass
+from datetime import datetime
+from typing import Self
+
+# pylint: disable=no-name-in-module
+from frequenz.api.common.v1.metrics.bounds_pb2 import Bounds
+from frequenz.api.common.v1.metrics.metric_sample_pb2 import (
+    AggregatedMetricSample as PBAggregatedMetricSample,
+)
+from frequenz.api.common.v1.metrics.metric_sample_pb2 import Metric as PBMetric
+from frequenz.api.common.v1.metrics.metric_sample_pb2 import (
+    MetricSample as PBMetricSample,
+)
+from frequenz.api.common.v1.metrics.metric_sample_pb2 import (
+    MetricSampleVariant as PBMetricSampleVariant,
+)
+from frequenz.api.common.v1.metrics.metric_sample_pb2 import (
+    SimpleMetricSample as PBSimpleMetricSample,
+)
+from google.protobuf.timestamp_pb2 import Timestamp
+
+# pylint: enable=no-name-in-module
+
+
+class Metric(enum.Enum):
+    """List of supported metrics."""
+
+    # Default value
+    UNSPECIFIED = PBMetric.METRIC_UNSPECIFIED
+
+    # DC electricity metrics
+    DC_VOLTAGE = PBMetric.METRIC_DC_VOLTAGE
+    DC_CURRENT = PBMetric.METRIC_DC_CURRENT
+    DC_POWER = PBMetric.METRIC_DC_POWER
+
+    # General AC electricity metrics
+    AC_FREQUENCY = PBMetric.METRIC_AC_FREQUENCY
+    AC_VOLTAGE = PBMetric.METRIC_AC_VOLTAGE
+    AC_VOLTAGE_PHASE_1 = PBMetric.METRIC_AC_VOLTAGE_PHASE_1
+    AC_VOLTAGE_PHASE_2 = PBMetric.METRIC_AC_VOLTAGE_PHASE_2
+    AC_VOLTAGE_PHASE_3 = PBMetric.METRIC_AC_VOLTAGE_PHASE_3
+    AC_APPARENT_CURRENT = PBMetric.METRIC_AC_APPARENT_CURRENT
+    AC_APPARENT_CURRENT_PHASE_1 = PBMetric.METRIC_AC_APPARENT_CURRENT_PHASE_1
+    AC_APPARENT_CURRENT_PHASE_2 = PBMetric.METRIC_AC_APPARENT_CURRENT_PHASE_2
+    AC_APPARENT_CURRENT_PHASE_3 = PBMetric.METRIC_AC_APPARENT_CURRENT_PHASE_3
+
+    # AC power metrics
+    AC_APPARENT_POWER = PBMetric.METRIC_AC_APPARENT_POWER
+    AC_APPARENT_POWER_PHASE_1 = PBMetric.METRIC_AC_APPARENT_POWER_PHASE_1
+    AC_APPARENT_POWER_PHASE_2 = PBMetric.METRIC_AC_APPARENT_POWER_PHASE_2
+    AC_APPARENT_POWER_PHASE_3 = PBMetric.METRIC_AC_APPARENT_POWER_PHASE_3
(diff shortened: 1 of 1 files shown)


Marenz:
Seems to need a reformatting with `black` so the CI is happy. 
And either a release notes entry or the label to skip that test

flora-hofmann-frequenz:
> Seems to need a reformatting with `black` so the CI is happy. And either a release notes entry or the label to skip that test

I've done a `black` & `flake8` before uploading. strange...

Marenz:
Oh and, the DCO check is failing because .. well it's better explained here: https://github.com/frequenz-floss/frequenz-client-common-python/pull/8/checks?check_run_id=21060093415

Not sure if that implies or requires them to be PGP signed as well.

src/frequenz/client/common/_types.py
```diff
@@ -0,0 +1,1022 @@
...
+    pagination_params_pb2,
+    pagination_info_pb2,
+)
+from google.protobuf import timestamp_pb2
+
+# Set up logging
+_logger = logging.getLogger(__name__)
+
+
+# # From common metric api
```
llucax: Nitpick, but any reason for the double `#`?

src/frequenz/client/common/_types.py
```diff
@@ -0,0 +1,1022 @@
...
+class Metric(enum.Enum):
+    """
+    List of supported metrics.
+    """
+
+    # Default value
+    UNSPECIFIED = metric_sample_pb2.Metric.UNSPECIFIED
+
+    # DC electricity metrics
+    DC_VOLTAGE = metric_sample_pb2.Metric.METRIC_DC_VOLTAGE
```
llucax: Here in general the documentation from the protobuf file should be copied. I can see here how automatic generation of enums could be a huge time saver.

We don't have to do it for this first PR though.

llucax (commented):
I just made a couple of very generic comments, and then see that this doesn't have many of the things we mentioned are present in the SDK (like `MeterData`, etc.), so I guess this is just a raw dump of what you did in reporting, to get things started, but there will be follow up PRs. Is this correct?

cwasicki:
> so I guess this is just a raw dump of what you did in reporting, to get things started, but there will be follow up PRs. Is this correct?

I think so, in my opinion having this in separate PRs is cleaner.

flora-hofmann-frequenz:
> I just made a couple of very generic comments, and then see that this doesn't have many of the things we mentioned are present in the SDK (like `MeterData`, etc.), so I guess this is just a raw dump of what you did in reporting, to get things started, but there will be follow up PRs. Is this correct?

Yes, that's correct. I only added what was in the reporting API PR so far (and changed the List error).

llucax:
Yup, sounds good to me.

src/frequenz/client/common/_types.py
```diff
@@ -0,0 +1,1022 @@
...
+
+from __future__ import annotations  # required for constructor type hinting
+
+import enum
+import logging
+from datetime import datetime
+from dataclasses import dataclass
+from typing import List
+
+from frequenz.api.common.v1.metrics import metric_sample_pb2, bounds_pb2
```
cwasicki: Discussed this in today's meeting and decided to not go with this simplified approach, where we have one single module for all types. Instead we want to keep the protobuf structure, i.e. have separate `metrics`, `microgrid` and `pagination` modules.

src/frequenz/client/common/_types.py
flora-hofmann-frequenz: Will take it out when I split the module into separate files.

llucax:
Hi @flora-hofmann-frequenz , what's the status of this PR, I noticed you made some pushes recently but there are no request to re-review or any comments, so I'm not sure if you are still working on it or if it's ready for another round of review. In any case, please let us know if you need new feedback, ideally by clicking on the request review button.

image

cwasicki:
@flora-hofmann-frequenz  According to https://github.com/frequenz-floss/frequenz-client-common-python/pull/11/files the suggested structure would be to move everything into `__init__.py`

flora-hofmann-frequenz:
Still have 3 mypy errors that I can't quite figure out - can someone help with them pls?

shsms:
Sorry, a bit late to review, but do we really need these types from a client pov?

A client would need to stream component data and read values easily.  For that we need specific types without the nested `metric*` stuff inside, but just fields like power and voltage that have float values.

I think for these types, we should work backwards - start from what is useful for the end user and implement the types required for that.  Specifically thinking of types like this:  https://github.com/frequenz-floss/frequenz-sdk-python/blob/d34512ee0a95b91d717b65495509063eac9f0ab6/src/frequenz/sdk/microgrid/component/_component_data.py#L62-L120

flora-hofmann-frequenz:
> Sorry, a bit late to review, but do we really need these types from a client pov?

Not quite sure I follow this, does this mean I should scrap this PR?

Also have not worked with the SDK types before and would appreciate if someone with experience from the SDK would take the lead here then.

shsms:
> Not quite sure I follow this, does this mean I should scrap this PR?

That's my feeling, because ultimately what we need is to get battery data or inverter data.  The abstract and recursive types the API uses to represent this data is not something the application developers are interested in.  And if we take this in, we'll either drop it later without using it,  or we'll create unnecessary dependencies on code that didn't have to be there in the first place.

Maybe we get on a call tomorrow @flora-hofmann-frequenz? and do some pair programming to identify the minimum amount of types we need to have a stable external interface for the clients.

Marenz:
@shsms My understanding was, that all those types are requirements for the actors/clients that are used in the reporting-api.
 I also didn't think that this repo is related to what any application dev. needs but for the clients/actors. Only those are what the app devs use eventually, no?

shsms:
>  that all those types are requirements for the actors/clients that are used in the reporting-api.

Yes, but the clients are supposed to be higher level, and so should expose only strong types I think.

The API exposes all kinds of data as a list of kv pairs.  But the clients shouldn't do that as well,  and ask users of the client to loop through the kv pairs and find the metrics they want.  The client should provide ways to access the data a lot more easily, like attributes in a class.

If the clients don't expose kv pairs,  then we don't need the `Metric` type, for example from this PR.

Marenz:
What does this repo have to do with the types that any client exposes? The client needs them, not the user, even if only internally?

shsms:
> The client needs them, not the user, even if only internally?

If I'm understanding this correctly, the clients themselves don't need wrapper types, because they already know about the protobuf types and can use those directly.

The clients should hide the protobuf types from the users of the client.  We don't have to protect the reporting or microgrid clients from the raw protobuf types found in common.

llucax:
> If I'm understanding this correctly, the clients themselves don't need wrapper types, because they already know about the protobuf types and can use those directly.
> 
> The clients should hide the protobuf types from the users of the client. We don't have to protect the reporting or microgrid clients from the raw protobuf types found in common.

Meh, Maybe wrapping could have some benefits, like doing validation. We have the same *issue* with the pagination stuff, that should never be exposed to client **users**, but it might be useful for client **developers**.

But I agree it seems we are going a bit backwards, and if what's happening here is just writing wrappers mainly as a 1-1 mapping to protobuf stuff, then it doesn't make a lot of sense and we should first write the clients and then see what's worth sharing and move it here.

@flora-hofmann-frequenz I'm not sure how urgently you need this stuff for reporting, but I will be getting involved very soon in the microgrid API which share a lot of the types, if you still have some other work to do, maybe we can sync about this and how to get stuff that is useful for both reporting and microgrid when I'm more involved in the topic.

shsms:
> Maybe wrapping could have some benefits, like doing validation

we should do validation directly on the protobuf objects, or the final objects created for the clients.  else we're just wasting cpu cycles converting from one iterative type to another, and then iterate again to create the final struct.

llucax:
Yeah, I agree that just a validation function would be enough in those cases.

flora-hofmann-frequenz:
Closed as per recent discussion, instead working on metric PR #21 to work with mvp of reporting client.

PR #11, opened 2024-02-06
Add component types

Marenz:
~Depends on #10~

- Delete cookiecutter template leftovers
- Add components types from SDK

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,6 @@ description = "Common code and utilities for Frequenz API clients"
 readme = "README.md"
 license = { text = "MIT" }
 keywords = ["frequenz", "python", "lib", "library", "client-common"]
-# TODO(cookiecutter): Remove and add more classifiers if appropriate
 classifiers = [
   "Development Status :: 3 - Alpha",
   "Intended Audience :: Developers",
@@ -26,7 +25,6 @@ classifiers = [
   "Typing :: Typed",
 ]
 requires-python = ">= 3.11, < 4"
-# TODO(cookiecutter): Remove and add more dependencies if appropriate
 dependencies = [
   "typing-extensions >= 4.5.0, < 5",
 ]
@@ -36,7 +34,6 @@ dynamic = ["version"]
 name = "Frequenz Energy-as-a-Service GmbH"
 email = "floss@frequenz.com"
 
-# TODO(cookiecutter): Remove and add more optional dependencies if appropriate
 [project.optional-dependencies]
 dev-flake8 = [
   "flake8 == 6.1.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,7 @@ classifiers = [
 requires-python = ">= 3.11, < 4"
 dependencies = [
   "typing-extensions >= 4.5.0, < 5",
+  "frequenz-api-common >= 0.5.4, < 6",
 ]
 dynamic = ["version"]
 
--- a/src/frequenz/client/common/microgrid/__init__.py
+++ b/src/frequenz/client/common/microgrid/__init__.py
@@ -1,0 +1,4 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Frequenz microgrid definition."""
--- a/src/frequenz/client/common/__init__.py
+++ b/src/frequenz/client/common/__init__.py
@@ -1,25 +1,4 @@
 # License: MIT
 # Copyright © 2023 Frequenz Energy-as-a-Service GmbH
 
-"""Common code and utilities for Frequenz API clients.
-
-TODO(cookiecutter): Add a more descriptive module description.
-"""
-
-
-# TODO(cookiecutter): Remove this function
-def delete_me(*, blow_up: bool = False) -> bool:
-    """Do stuff for demonstration purposes.
-
-    Args:
-        blow_up: If True, raise an exception.
-
-    Returns:
-        True if no exception was raised.
-
-    Raises:
-        RuntimeError: if blow_up is True.
-    """
-    if blow_up:
-        raise RuntimeError("This function should be removed!")
-    return True
+"""Common code and utilities for Frequenz API clients."""
--- a/src/frequenz/client/common/microgrid/components/__init__.py
+++ b/src/frequenz/client/common/microgrid/components/__init__.py
@@ -1,0 +1,67 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+"""Defines the components that can be used in a microgrid."""
+from __future__ import annotations
+
+from enum import Enum
+
+# pylint: disable=no-name-in-module
+from frequenz.api.common.v1.microgrid.components.components_pb2 import (
+    ComponentCategory as PBComponentCategory,
+)
+
+# pylint: enable=no-name-in-module
+
+
+class ComponentCategory(Enum):
+    """Possible types of microgrid component."""
+
+    UNSPECIFIED = PBComponentCategory.COMPONENT_CATEGORY_UNSPECIFIED
+    """An unknown component category.
+
+    Useful for error handling, and marking unknown components in
+    a list of components with otherwise known categories.
+    """
+
+    GRID = PBComponentCategory.COMPONENT_CATEGORY_GRID
+    """The point where the local microgrid is connected to the grid."""
+
+    METER = PBComponentCategory.COMPONENT_CATEGORY_METER
+    """A meter, for measuring electrical metrics, e.g., current, voltage, etc."""
+
+    INVERTER = PBComponentCategory.COMPONENT_CATEGORY_INVERTER
+    """An electricity generator, with batteries or solar energy."""
+
+    BATTERY = PBComponentCategory.COMPONENT_CATEGORY_BATTERY
+    """A storage system for electrical energy, used by inverters."""
+
+    EV_CHARGER = PBComponentCategory.COMPONENT_CATEGORY_EV_CHARGER
+    """A station for charging electrical vehicles."""
+
+    CHP = PBComponentCategory.COMPONENT_CATEGORY_CHP
+    """A heat and power combustion plant (CHP stands for combined heat and power)."""
+
+    @classmethod
+    def from_proto(
+        cls, component_category: PBComponentCategory.ValueType
+    ) -> ComponentCategory:
+        """Convert a protobuf ComponentCategory message to ComponentCategory enum.
+
+        Args:
+            component_category: protobuf enum to convert
+
+        Returns:
+            Enum value corresponding to the protobuf message.
+        """
+        if not any(t.value == component_category for t in ComponentCategory):
+            return ComponentCategory.UNSPECIFIED
+        return cls(component_category)
+
+    def to_proto(self) -> PBComponentCategory.ValueType:
+        """Convert a ComponentCategory enum to protobuf ComponentCategory message.
+
+        Returns:
+            Enum value corresponding to the protobuf message.
+        """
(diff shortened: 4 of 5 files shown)


pyproject.toml
```diff
@@ -26,17 +25,16 @@ classifiers = [
   "Typing :: Typed",
 ]
 requires-python = ">= 3.11, < 4"
-# TODO(cookiecutter): Remove and add more dependencies if appropriate
 dependencies = [
   "typing-extensions >= 4.5.0, < 5",
+  "frequenz-api-microgrid >= 0.15.3, < 0.16.0",
```
llucax: This cannot depend on the `microgrid-api`, the dependency should be the other way around.

src/frequenz/client/common/_component.py
llucax, line 1: I would make the module public instead, all the entities in the different directories in the protobuf are pretty much independent from each other.

I think also for this particular repo it would be very useful to stay as close as possible to the `api-common` repository, so I would rename this file as: `src/frequenz/client/common/microgrid/components/__init__.py` and put there everything related to `proto/frequenz/api/common/v1/microgrid/components/components.proto` there, and create one public file per any other `proto/frequenz/api/common/v1/microgrid/components/*.py` inside it, for example:

`proto/frequenz/api/common/v1/microgrid/components/battery.proto `src/frequenz/client/common/microgrid/components/battery.py`.

I would probably even go as far as including the version in the python path too, as I don't think it would be crazy to think that a client might need to support with multiple versions at the same time, so `src/frequenz/client/common/v1/microgrid/components/__init__.py` and `src/frequenz/client/common/v1/microgrid/components/battery.py` for example.

I would not re-export symbols in `src/frequenz/client/common/__init__.py`, as this repo has many unrelated stuff I think it is better if uses just use the full path when importing.

src/frequenz/client/common/_component.py
```diff
@@ -0,0 +1,204 @@
...
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+"""Defines the components that can be used in a microgrid."""
+from __future__ import annotations
+
+from dataclasses import dataclass
+from enum import Enum
+
+import frequenz.api.common.components_pb2 as components_pb
+import frequenz.api.microgrid.inverter_pb2 as inverter_pb
```
llucax: I guess you need the microgrid API because you are targeting the microgrid API v0.15, in v0.16 this was [moved to `api-common`](https://github.com/frequenz-floss/frequenz-api-common/tree/v0.x.x/proto/frequenz/api/common/v1/microgrid/components).

All the new code, specially in common, should use the microgrid API v0.16, otherwise we'll get into a big mess.

src/frequenz/client/common/_component.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+class ComponentType(Enum):
+    """A base class from which individual component types are derived."""
+
+
+# pylint: disable=no-member
```
llucax: It is better to use this scheme:

```py
import frequenz.api.common.components_pb2 import ComponentCategory 
```

Otherwise mistake trying to use non-existing names in a module will not be caught by `mypy` and will appear only in runtime.

For example: https://github.com/frequenz-floss/frequenz-sdk-python/blob/91821a3e2719024c0b2a114d6a11aa692d3429d6/src/frequenz/sdk/actor/power_distributing/_component_status/_battery_status_tracker.py#L12-L18

FYI @frequenz-floss/api-team

src/frequenz/client/common/_component.py
```diff
@@ -0,0 +1,204 @@
...
+    def __hash__(self) -> int:
+        """Compute a hash of this instance, obtained by hashing the `component_id` field.
+
+        Returns:
+            Hash of this instance.
+        """
+        return hash(self.component_id)
+
+
+class ComponentMetricId(Enum):
```
llucax: Are these defined anywhere in the `api-common` repo? Or where do they come from?

llucax (commented):
This probably need a better structure, closer to the API definitions structure, but like with #8 I'm OK to merge it if the idea is just to get things moving an improve in followup PRs.

The dependency to microgrid-api is kind of big an problematic though, so I probably would not merge that.

src/frequenz/client/common/_component.py
Marenz: Straight from the SDK...

Marenz: What I am trying to say: I didn't check where anything came from, I just copied the file and removed `Fuse`-dependent stuff

src/frequenz/client/common/_component.py
Marenz: should I be using .v1 in those import paths then?

llucax: I guess you mean `/v1`? If so, yes.

src/frequenz/client/common/_component.py
llucax: OK, again, if we want to first commit the code copied from the SDK as is, then great, we can merge as is, but I would make it very clear in the commit message and PR description that this is the intention.

Marenz: Nah, fuck that, I'll do it properly from the beginning

Marenz: Though, import paths like `from frequenz.client.common.microgrid.components.components import ComponentCategory` still make me doubt that recreating that structure is the right way

Marenz:
I removed now everything that I don't really need for dispatch and put it in the right place in the structure.

src/frequenz/client/common/_component.py
Marenz: I'd abstract away the version here because it's easy to support multiple on this level, compared to the protobuf level

llucax: Yeah, true. At some point we'll need to make breaking changes at this level too, but that's a broader discussion if we want to do something like using `vX` packages to allow users use multiple incompatible versions in the same code. I'm really starting to believe that could be a good idea in general... But again, separate discussion.

src/frequenz/client/common/_component.py
llucax: No, I think `components/components.proto` should be translated to `components/__init__.py` in here, so it should be `frequenz.client.common.microgrid.components import ComponentCategory`.

src/frequenz/client/common/microgrid/__init__.py
```diff
@@ -0,0 +1,4 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Common code and utilities for Frequenz API clients."""
```
llucax: Maybe copy the module-level docs from https://github.com/frequenz-floss/frequenz-api-common/blob/v0.x.x/proto/frequenz/api/common/v1/microgrid/microgrid.proto?

It doesn't seem to be very useful to copy the same generic string on every module.

src/frequenz/client/common/microgrid/components/components.py
llucax, line 1: -> `src/frequenz/client/common/microgrid/components/__init__.py`

src/frequenz/client/common/microgrid/components/components.py
```diff
@@ -0,0 +1,74 @@
...
+)
+
+# pylint: enable=no-name-in-module
+
+
+class ComponentCategory(Enum):
+    """Possible types of microgrid component."""
+
+    NONE = PBComponentCategory.COMPONENT_CATEGORY_UNSPECIFIED
+    """Unspecified component category."""
```
llucax: Same about the documentation of all these items, we should use the same as in the protobuf stuff, or at least that at the minimum/base, maybe we can extend it with python-specific stuff too (or remove stuff that it is too protobuf-specific).

src/frequenz/client/common/microgrid/components/components.py
```diff
@@ -0,0 +1,74 @@
...
+    """Battery component."""
+
+    EV_CHARGER = PBComponentCategory.COMPONENT_CATEGORY_EV_CHARGER
+    """EV charger component."""
+
+    CHP = PBComponentCategory.COMPONENT_CATEGORY_CHP
+    """CHP component."""
+
+
+def _component_category_from_protobuf(
```
llucax: Since we are starting from scratch here, I would add a `@classmethod` `from_proto()` to the `enum` as we do with classes.

I'm not sure if I mentioned this, but I was thinking that if the idea is not to expose the protobuf stuff, when it might be better to have the to/from proto conversion separate (even in a separate module, so protobuf stuff is not imported at all when only using the wrappers), but then for enums that would be super inconvenient because we wouldn't be able to use the same values as in the protobuf stuff, which could be very useful when debugging (or we'll need to duplicate everything manually), so I guess for a first version it might make more sense to keep it simple (and coupled).

src/frequenz/client/common/microgrid/__init__.py
Marenz: The proto module-level docs seem even more generic...

```
// Frequenz microgrid definition.
//
// Copyright 2023 Frequenz Energy-as-a-Service GmbH
//
// Licensed under the MIT License (the "License");
// you may not use this file except in compliance with the License.
```

Marenz: to be fair, I think none of these init files seem very useful to me ;)

src/frequenz/client/common/_component.py
Marenz: And how would the other proto files then be translated?

```
components/
  battery.proto
  components.proto
  ev_charger.proto
  fuse.proto
  grid.proto
  inverter.proto
  transformer.proto
```

Marenz:
Added more docu, made to/from proto methods members.

src/frequenz/client/common/_component.py
llucax: `components/battery.py`, etc. I think we just treat `xxx/xxx.proto` specially and have it as `xxx/__init__.py` instead.

llucax: As far as I see this pattern is used in other dirs too, for example `microgrid/microgrid.proto`, and usually the `xxx/xxx.proto` file uses other `xxx/yyy.proto` files and not the other way around.

@tiyash-basu-frequenz FYI, just to confirm that his pattern is used consciously and not just a coincidence. It would be important to have a consistent scheme to make the python modules make sense and make sure we don't run into cyclic dependencies.

In the case of `microgrid/sensor/sensor.proto`, where there is only that file in the `microgrid/sensor/` directory, I would just add that file as a plain `microgrid/sensor.py`. If more files are added in the future we can rename `sensor.py` to sensor/__init__.py` and it will be transparent to the end user.

src/frequenz/client/common/microgrid/__init__.py
llucax: > to be fair, I think none of these init files seem very useful to me ;)

I agree on this one

> The proto module-level docs seem even more generic...

But the one in the proto at least has *microgrid* in it :D

Anyway, if in this case in the proto file there is nothing useful, all good, my main point is the documentation in here has to be at least as "good" as the one in the proto files.

Marenz:
Moved main types to __init__ as requested

src/frequenz/client/common/_component.py
tiyash-basu-frequenz: This is a pattern we consciously use (for protobuf), and this is a result of how we want to define packages in our protobuf definitions. `microgrid/sensor/sensor.proto` means that the entities in `sensor.proto` are a member of the package `...microgrid.sensor`. The same applies to all other protobuf files.

For references, check these links: https://github.com/frequenz-floss/frequenz-api-common/pull/127#discussion_r1385230614, https://github.com/frequenz-floss/frequenz-api-common/pull/131

llucax (approved):

PR #14, opened 2024-02-08
Add microgrid types

flora-hofmann-frequenz:
Split up of types modules to reflect protobuf structure.

### DIFF
--- a/src/frequenz/client/common/microgrid/__init__.py
+++ b/src/frequenz/client/common/microgrid/__init__.py
@@ -21,7 +21,7 @@ class MicrogridComponentIDs:
     microgrid_id: int
     """The ID of the microgrid."""
 
-    component_ids: List[int]
+    component_ids: Sequence[int]
     """List of component IDs belonging to this microgrid."""
 
     @classmethod
--- a/src/frequenz/client/common/microgrid/__init__.py
+++ b/src/frequenz/client/common/microgrid/__init__.py
@@ -8,7 +8,7 @@ from __future__ import annotations  # required for constructor type hinting
 from dataclasses import dataclass
 from typing import Self
 
-# pylint: disable=import-error, no-name-in-module
+# pylint: disable=no-name-in-module
 from frequenz.api.common.v1.microgrid.microgrid_pb2 import (
     MicrogridComponentIDs as PBMicrogridComponentIDs,
 )
--- a/src/frequenz/client/common/microgrid/__init__.py
+++ b/src/frequenz/client/common/microgrid/__init__.py
@@ -1,4 +1,51 @@
 # License: MIT
-# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
 
-"""Frequenz microgrid definition."""
+"""Module to define the microgrid used with the common client."""
+
+from __future__ import annotations  # required for constructor type hinting
+
+from dataclasses import dataclass
+from typing import List, Self
+
+# pylint: disable=import-error, no-name-in-module
+from frequenz.api.common.v1.microgrid.microgrid_pb2 import (
+    MicrogridComponentIDs as PBMicrogridComponentIDs,
+)
+
+
+@dataclass(frozen=True, kw_only=True)
+class MicrogridComponentIDs:
+    """Linking component IDs with their respective microgrid ID."""
+
+    microgrid_id: int
+    """The ID of the microgrid."""
+
+    component_ids: List[int]
+    """List of component IDs belonging to this microgrid."""
+
+    @classmethod
+    def from_proto(cls, microgrid_component_ids: PBMicrogridComponentIDs) -> Self:
+        """Convert a protobuf MicrogridComponentIDs to MicrogridComponentIDs object.
+
+        Args:
+            microgrid_component_ids: MicrogridComponentIDs to convert.
+
+        Returns:
+            MicrogridComponentIDs object corresponding to the protobuf message.
+        """
+        return cls(
+            microgrid_id=microgrid_component_ids.microgrid_id,
+            component_ids=List(microgrid_component_ids.component_ids),
+        )
+
+    def to_proto(self) -> PBMicrogridComponentIDs:
+        """Convert a MicrogridComponentIDs object to protobuf MicrogridComponentIDs.
+
+        Returns:
+            Protobuf message corresponding to the MicrogridComponentIDs object.
+        """
+        return PBMicrogridComponentIDs(
+            microgrid_id=self.microgrid_id,
+            component_ids=List(self.component_ids),
+        )
--- a/src/frequenz/client/common/microgrid/__init__.py
+++ b/src/frequenz/client/common/microgrid/__init__.py
@@ -12,7 +12,7 @@ from typing import Self
 from frequenz.api.common.v1.microgrid.microgrid_pb2 import (
     MicrogridComponentIDs as PBMicrogridComponentIDs,
 )
-
+# pylint: enable=no-name-in-module
 
 @dataclass(frozen=True, kw_only=True)
 class MicrogridComponentIDs:
--- a/src/frequenz/client/common/microgrid/__init__.py
+++ b/src/frequenz/client/common/microgrid/__init__.py
@@ -6,7 +6,7 @@
 from __future__ import annotations  # required for constructor type hinting
 
 from dataclasses import dataclass
-from typing import List, Self
+from typing import Self
 
 # pylint: disable=import-error, no-name-in-module
 from frequenz.api.common.v1.microgrid.microgrid_pb2 import (


src/frequenz/client/common/microgrid/microgrid.py
```diff
@@ -0,0 +1,59 @@
...
+from dataclasses import dataclass
+from typing import List
+
+# pylint: disable=import-error, no-name-in-module
+from frequenz.api.common.v1.microgrid.microgrid_pb2 import (
+    MicrogridComponentIDs as PBMicrogridComponentIDs,
+)
+
+# Set up logging
+_logger = logging.getLogger(__name__)
```
Marenz: Is logging needed / useful / used in this module?

src/frequenz/client/common/microgrid/microgrid.py
```diff
@@ -0,0 +1,59 @@
...
+
+"""Module to define the microgrid used with the common client."""
+
+from __future__ import annotations  # required for constructor type hinting
+
+import logging
+from dataclasses import dataclass
+from typing import List
+
+# pylint: disable=import-error, no-name-in-module
```
Marenz: I think `no-name-in-module` should be enough here. Also we want to enable it again after the import with

`# pylint: enable=no-name-in-module`

the CI error is related to mypy, not sure whats up there

src/frequenz/client/common/microgrid/microgrid.py
```diff
@@ -0,0 +1,59 @@
...
+    component_ids: List[int]
+
+    @classmethod
+    def from_pb(
+        cls, microgrid_component_ids: PBMicrogridComponentIDs
+    ) -> MicrogridComponentIDs:
+        """Convert a protobuf MicrogridComponentIDs to MicrogridComponentIDs object.
+
+        Args:
+            microgrid_component_ids: MicrogridComponentIDs to convert.
```
Marenz: missing blank line before `Returns:`

llucax:
I made some similar comments in:
* https://github.com/frequenz-floss/frequenz-client-common-python/pull/15

I think we should focus on the review of only one and delay the review of the other after that, because I think there will be a lot of duplicate comments.

Since I made quite a few comments to #15 and most comments done here are also covered there, I suggest to start with #15. I will make this a draft PR to make this clear. If you don't agree please feel free to change it back to a final PR.

src/frequenz/client/common/microgrid/__init__.py
```diff
@@ -1,4 +1,51 @@
 # License: MIT
-# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
 
-"""Frequenz microgrid definition."""
+"""Module to define the microgrid used with the common client."""
+
+from __future__ import annotations  # required for constructor type hinting
```
llucax, line 6: This shouldn't be necessary anymore if you use `Self`, right?

src/frequenz/client/common/microgrid/__init__.py
```diff
@@ -1,4 +1,51 @@
...
+
+from dataclasses import dataclass
+from typing import List, Self
```
llucax: Just use the built-in `list` type for type hints, `List` is deprecated.

```suggestion
from typing import Self
```

src/frequenz/client/common/microgrid/__init__.py
```diff
@@ -1,4 +1,51 @@
...
+
+# pylint: disable=import-error, no-name-in-module
```
llucax:
```suggestion
# pylint: disable=no-name-in-module
```

src/frequenz/client/common/microgrid/__init__.py
```diff
@@ -1,4 +1,51 @@
...
+from frequenz.api.common.v1.microgrid.microgrid_pb2 import (
+    MicrogridComponentIDs as PBMicrogridComponentIDs,
+)
+
```
llucax: Even when is the last import, I would still re-enable the check, just in case more imports are added in the future.

```suggestion
# pylint: enable=no-name-in-module
```

src/frequenz/client/common/microgrid/__init__.py
```diff
@@ -1,4 +1,51 @@
...
+
+@dataclass(frozen=True, kw_only=True)
+class MicrogridComponentIDs:
+    """Linking component IDs with their respective microgrid ID."""
+
+    microgrid_id: int
+    """The ID of the microgrid."""
+
+    component_ids: List[int]
```
llucax: Better use [`Sequence[int]`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence) here, as the class should be immutable. Otherwise people is not able to do `instance.component_ids = [1,2]` but they can still do `instance.component_ids.clear()` and `instance.component_ids.append(1)` for example.

```suggestion
    component_ids: Sequence[int]
```

src/frequenz/client/common/microgrid/__init__.py
```diff
@@ -1,4 +1,51 @@
...
+
+        Args:
+            microgrid_component_ids: MicrogridComponentIDs to convert.
+
+        Returns:
+            MicrogridComponentIDs object corresponding to the protobuf message.
+        """
+        return cls(
+            microgrid_id=microgrid_component_ids.microgrid_id,
+            component_ids=List(microgrid_component_ids.component_ids),
```
llucax, line 39: This raises an interesting question, copying the list (or whatever `component_ids` is in the proto class) is definitely the safe option, but it means doing some copying, when in 99% of the cases you'll probably just take the protobuf message, convert it to the wrapper and discard it, so it is very unlikely it will be mutated after creating the wrapper (in which case it will also affect the wrapper list, as we stored a reference to the protobuf message).

Maybe since it is unlikely and might have a performance impact, we could document that a reference is taken and the users should take care of the copying if they plan to keep mutating the proto message instance.

@frequenz-floss/python-sdk-team any opinions?

src/frequenz/client/common/microgrid/__init__.py
```diff
@@ -1,4 +1,51 @@
...
+    def to_proto(self) -> PBMicrogridComponentIDs:
+        """Convert a MicrogridComponentIDs object to protobuf MicrogridComponentIDs.
+
+        Returns:
+            Protobuf message corresponding to the MicrogridComponentIDs object.
+        """
+        return PBMicrogridComponentIDs(
+            microgrid_id=self.microgrid_id,
+            component_ids=List(self.component_ids),
+        )
```
llucax, line 51: Same here, if the common case would just be: `client.send(instance.to_proto())`, then copying doesn't make a lot of sense.

src/frequenz/client/common/microgrid/__init__.py
shsms: Haven't really figured out what this type is for, but it still looks like a one time thing that we might have to process at startup, and not along with each streaming message, etc.  So I don't see any benefit from not choosing the conservative approach.

llucax: Yeah, this is a more general question, because we will probably have it in other situations, as for this particular class, I guess we need to figure out if we want to pursue this PR as you mentioned in https://github.com/frequenz-floss/frequenz-client-common-python/pull/8#issuecomment-1954809138, I think it was a good point that we might be missing the end goal here.

src/frequenz/client/common/microgrid/__init__.py
```diff
@@ -1,4 +1,51 @@
...
+# pylint: enable=no-name-in-module
+
+@dataclass(frozen=True, kw_only=True)
+class MicrogridComponentIDs:
+    """Linking component IDs with their respective microgrid ID."""
+
+    microgrid_id: int
+    """The ID of the microgrid."""
+
+    component_ids: Sequence[int]
```
daniel-zullo-frequenz, line 24: `Sequence` was not imported

src/frequenz/client/common/microgrid/__init__.py
daniel-zullo-frequenz: I agree it is fine to copy the IDs if the message is only processed once at startup (or only a few times). For general scenarios it makes sense to avoid the copy if it is possible

src/frequenz/client/common/microgrid/__init__.py
daniel-zullo-frequenz: Actually the type here might depend on the resolution of https://github.com/frequenz-floss/frequenz-client-common-python/pull/14/files#r1495647444

src/frequenz/client/common/microgrid/__init__.py
llucax: And to @shsms point, if this is known to never leave the client, then we always know the protobuf is discarded and we don't need to copy. All points to not really having to copy stuff when converting from/to protobuf in this repo, again under the premise that types here will never be exposed to the user as is.

flora-hofmann-frequenz:
Closed as this will also be changed in connection with reporting client mvp.

PR #15, opened 2024-02-08
Add pagination types

flora-hofmann-frequenz:
Split up type modules to reflect protobuf structure.

### DIFF
--- a/src/frequenz/client/common/pagination/__init__.py
+++ b/src/frequenz/client/common/pagination/__init__.py
@@ -1,0 +1,87 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Module to define the pagination used with the common client."""
+
+from __future__ import annotations  # required for constructor type hinting
+
+from dataclasses import dataclass
+from typing import Self
+
+# pylint: disable=no-name-in-module
+from frequenz.api.common.v1.pagination.pagination_info_pb2 import PaginationInfo
+from frequenz.api.common.v1.pagination.pagination_params_pb2 import PaginationParams
+
+# pylint: enable=no-name-in-module
+
+
+@dataclass(frozen=True, kw_only=True)
+class Params:
+    """Parameters for paginating list requests."""
+
+    page_size: int | None = None
+    """The maximum number of results to be returned per request."""
+
+    page_token: str | None = None
+    """The token identifying a specific page of the list results."""
+
+    @classmethod
+    def from_proto(cls, pagination_params: PaginationParams) -> Self:
+        """Convert a protobuf Params to PaginationParams object.
+
+        Args:
+            pagination_params: Params to convert.
+        Returns:
+            Params object corresponding to the protobuf message.
+        """
+        return cls(
+            page_size=pagination_params.page_size,
+            page_token=pagination_params.page_token,
+        )
+
+    def to_proto(self) -> PaginationParams:
+        """Convert a Params object to protobuf PaginationParams.
+
+        Returns:
+            Protobuf message corresponding to the Params object.
+        """
+        return PaginationParams(
+            page_size=self.page_size,
+            page_token=self.page_token,
+        )
+
+
+@dataclass(frozen=True, kw_only=True)
+class Info:
+    """Information about the pagination of a list request."""
+
+    total_items: int
+    """The total number of items that match the request."""
+
+    next_page_token: str | None = None
+    """The token identifying the next page of results."""
+
+    @classmethod
+    def from_proto(cls, pagination_info: PaginationInfo) -> Self:
+        """Convert a protobuf PBPaginationInfo to Info object.
+
+        Args:
+            pagination_info: Info to convert.
+        Returns:
+            Info object corresponding to the protobuf message.
+        """
+        return cls(
+            total_items=pagination_info.total_items,
+            next_page_token=pagination_info.next_page_token,
+        )
+
+    def to_proto(self) -> PaginationInfo:
+        """Convert a Info object to protobuf PBPaginationInfo.
+
+        Returns:
+            Protobuf message corresponding to the Info object.
+        """
+        return PaginationInfo(
+            total_items=self.total_items,
+            next_page_token=self.next_page_token,
+        )


src/frequenz/client/common/pagination/__init__.py
```diff
@@ -0,0 +1,25 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Common code and utilities for Frequenz API clients.
+
+TODO(cookiecutter): Add a more descriptive module description.
+"""
```
llucax:
```suggestion
"""Common types to handle pagination of request results."""
```
Or something more along those lines?

src/frequenz/client/common/pagination/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+
+    Returns:
+        True if no exception was raised.
+
+    Raises:
+        RuntimeError: if blow_up is True.
+    """
+    if blow_up:
+        raise RuntimeError("This function should be removed!")
+    return True
```
llucax: As the comment suggest, this should be removed, this is only a code example in the repo-config template, but it shouldn't be included in all `__init__.py` files.

```suggestion
```

src/frequenz/client/common/pagination/__init__.py
```diff
@@ -0,0 +1,25 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
```
llucax:
```suggestion
# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
```

src/frequenz/client/common/pagination/pagination.py
llucax, line 1: As pointed out in https://github.com/frequenz-floss/frequenz-client-common-python/pull/11#discussion_r1482593600 I think we should move the contents of this file to `src/frequenz/client/common/pagination/__init____.py`.

src/frequenz/client/common/pagination/pagination.py
```diff
@@ -0,0 +1,100 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Module to define the pagination used with the common client."""
+
+from __future__ import annotations  # required for constructor type hinting
+
+import logging
```
llucax: I would remove this, as it is unused.

```suggestion
```

src/frequenz/client/common/pagination/pagination.py
```diff
@@ -0,0 +1,100 @@
...
+    PaginationInfo as PBPaginationInfo,
+)
+
+# pylint: disable=import-error, no-name-in-module
+from frequenz.api.common.v1.pagination_params_pb2 import (
+    PaginationParams as PBPaginationParams,
+)
+
+# Set up logging
+_logger = logging.getLogger(__name__)
```
llucax: Because this one is unused, and since it's a *private* member of the module (starts with `_`), this might probably fail a check in `flake8` or `pylint`.

```suggestion
```

src/frequenz/client/common/pagination/pagination.py
```diff
@@ -0,0 +1,100 @@
...
+
+# pylint: disable=import-error, no-name-in-module
+from frequenz.api.common.v1.pagination_info_pb2 import (
+    PaginationInfo as PBPaginationInfo,
+)
+
+# pylint: disable=import-error, no-name-in-module
+from frequenz.api.common.v1.pagination_params_pb2 import (
+    PaginationParams as PBPaginationParams,
+)
```
llucax: `# pylint: disable=import-error, no-name-in-module` in its own line disables that check from the point it appears until **the end of the file**. Also, I think on `no-name-in-module` is necessary. So you should do this instead:

```suggestion
# pylint: disable=no-name-in-module
from frequenz.api.common.v1.pagination_info_pb2 import (
    PaginationInfo as PBPaginationInfo,
)
from frequenz.api.common.v1.pagination_params_pb2 import (
    PaginationParams as PBPaginationParams,
)
# pylint: enable=no-name-in-module
```

src/frequenz/client/common/pagination/pagination.py
```diff
@@ -0,0 +1,100 @@
...
+
+# Set up logging
+_logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class PaginationParams:
```
llucax: I think here we need to make another design decision. Since in Python we already have the module name (`pagination`) maybe it is too redundant to make the class `PaginationParams` and we can just use `Params` to encourage the google-style imports `from frequenz.client.common import pagination; p = pagination.Params(...)`.

I think because of how proto files are structured, this makes sense.

FYI @frequenz-floss/python-sdk-team

llucax: Here comes another design question. Do we want a wrapper over `PaginationParams` at all? I envision this never being exposed to the user. The client should probably offer a method with an async iterable interface, like:

```py
async def list(self, page_size: int | None = None, page_token: str | None = None) -> AsyncIterator[Whatever]:
```

So users can do:

```py
async for whatever in client.list():
    ...
```

Then `list()` (or the async iterator returned by it) should handle pagination internally, pagination is basically a protocol optimization. Maybe there could be advanced cases where users want to use explicit pagination, but I think that should be a low level interface. Only if we want to offer this low-level interface I think it makes sense to wrap `PaginationParams`.

src/frequenz/client/common/pagination/pagination.py
```diff
@@ -0,0 +1,100 @@
...
+    Args:
+        page_size: The maximum number of results to be returned per request.
+        page_token: The token identifying a specific page of the list results.
+    """
+
+    page_size: int | None = None
+    page_token: str | None = None
+
+    @classmethod
+    def from_pb(cls, pagination_params: PBPaginationParams) -> PaginationParams:
```
llucax: I would use `from_proto()` which is what we use in the SDK, I don't see a clear advantage of using `from_pb` and if we change it it will be a lot more work to update the SDK. Also please use `Self` (`from typing import Self`) as the return type, otherwise this won't work correctly if someone wants to subclass this for some reason.

```suggestion
    def from_proto(cls, pagination_params: PBPaginationParams) -> Self:
```

src/frequenz/client/common/pagination/pagination.py
```diff
@@ -0,0 +1,100 @@
...
+            pagination_params: PaginationParams to convert.
+        Returns:
+            PaginationParams object corresponding to the protobuf message.
+        """
+        return cls(
+            page_size=pagination_params.page_size,
+            page_token=pagination_params.page_token,
+        )
+
+    def to_pb(self) -> PBPaginationParams:
```
llucax:
```suggestion
    def to_proto(self) -> PBPaginationParams:
```

src/frequenz/client/common/pagination/pagination.py
```diff
@@ -0,0 +1,100 @@
...
+# pylint: disable=import-error, no-name-in-module
+from frequenz.api.common.v1.pagination_params_pb2 import (
+    PaginationParams as PBPaginationParams,
+)
+
+# Set up logging
+_logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
```
llucax: I would use `kw_only` too here, to make the constructor more readable, as the attributes are not obvious if you write `Params(4)`, what is 4??? `Params(page_size=4)` looks more readable and obvious.

```suggestion
@dataclass(frozen=True, kw_only=True)
```

src/frequenz/client/common/pagination/pagination.py
```diff
@@ -0,0 +1,100 @@
...
+        Returns:
+            Protobuf message corresponding to the PaginationParams object.
+        """
+        return PBPaginationParams(
+            page_size=self.page_size,
+            page_token=self.page_token,
+        )
+
+
+@dataclass(frozen=True)
```
llucax:
```suggestion
@dataclass(frozen=True, kw_only)
```

src/frequenz/client/common/pagination/pagination.py
```diff
@@ -0,0 +1,100 @@
...
+    Args:
+        total_items: The total number of items that match the request.
+        next_page_token: The token identifying the next page of results.
+    """
+
+    total_items: int
+    next_page_token: str | None = None
+
+    @classmethod
+    def from_pb(cls, pagination_info: PBPaginationInfo) -> PaginationInfo:
```
llucax:
```suggestion
    def from_proto(cls, pagination_info: PBPaginationInfo) -> Self:
```

src/frequenz/client/common/pagination/pagination.py
```diff
@@ -0,0 +1,100 @@
...
+            pagination_info: PaginationInfo to convert.
+        Returns:
+            PaginationInfo object corresponding to the protobuf message.
+        """
+        return cls(
+            total_items=pagination_info.total_items,
+            next_page_token=pagination_info.next_page_token,
+        )
+
+    def to_pb(self) -> PBPaginationInfo:
```
llucax:
```suggestion
    def to_proto(self) -> PBPaginationInfo:
```

src/frequenz/client/common/pagination/pagination.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+    Information about the pagination of a list request.
+
+    Args:
+        total_items: The total number of items that match the request.
+        next_page_token: The token identifying the next page of results.
+    """
+
+    total_items: int
+    next_page_token: str | None = None
```
llucax: What happens when `next_page_token` is `None`?

```suggestion
    """Information about the pagination of a list request."""

    total_items: int
    """The total number of items that match the request."""

    next_page_token: str | None = None
    """The token identifying the next page of results."""
```

src/frequenz/client/common/pagination/pagination.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+    Parameters for paginating list requests.
+
+    Args:
+        page_size: The maximum number of results to be returned per request.
+        page_token: The token identifying a specific page of the list results.
+    """
+
+    page_size: int | None = None
+    page_token: str | None = None
```
llucax: What happens when any of them is `None`? I think we should document that too.

The documentation of attributes should be put below the attributes instead of in `Args` in the class, otherwise the docs will not render nicely

```suggestion
    """Parameters for paginating list requests."""

    page_size: int | None = None
    """The maximum number of results to be returned per request."""
    
    page_token: str | None = None
    """The token identifying a specific page of the list results."""
```

llucax (commented):
Hi Flora, I have a lot of comments, I hope it is not too overwhelming. They are actually a lot because I made a lot of suggestions so it is easier for you to update the code (if you accept them), without that probably it would have been one fourth of comments.

Most comments are minor, about style or naming, but there are a couple of bigger questions about design decisions, that I guess we need to make before we proceed adding all this stuff. Maybe we can discuss it in the next meeting.

src/frequenz/client/common/pagination/pagination.py
flora-hofmann-frequenz: Will do so after confirmation in yesterday's API discussion.

flora-hofmann-frequenz:
@cwasicki We have decided to start with this PR and then change #8 and #14 accordingly - see (https://github.com/frequenz-floss/frequenz-client-common-python/pull/14#issuecomment-1935550822).

src/frequenz/client/common/pagination/pagination.py
flora-hofmann-frequenz: I would rather move this to an Issue and then address it in a separate PR? Can we do that?

flora-hofmann-frequenz:
Maybe someone can help me to find out why `mypy` is still failing?

llucax:
> Maybe someone can help me to find out why `mypy` is still failing?

I think the import is incorrect:

>  src/frequenz/client/common/pagination/__init__.py:12: error: Skipping analyzing "frequenz.api.common.v1.pagination_info_pb2

Shouldn't it be `frequenz.api.common.v1.pagination.pagination_info_pb2`?

```py
>>> import frequenz.api.common.v1.pagination_info_pb2
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'frequenz.api.common.v1.pagination_info_pb2'
>>> import frequenz.api.common.v1.pagination.pagination_info_pb2
```

src/frequenz/client/common/pagination/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+"""Module to define the pagination used with the common client."""
+
+from __future__ import annotations  # required for constructor type hinting
+
+from dataclasses import dataclass
+from typing import Self
+
+# pylint: disable=no-name-in-module
+from frequenz.api.common.v1.pagination_info_pb2 import Info as PBPaginationInfo
+from frequenz.api.common.v1.pagination_params_pb2 import Params as PBPaginationParams
```
llucax: The imports seem to be messed up. It should be:

```suggestion
from frequenz.api.common.v1.pagination.pagination_info_pb2 import PaginationInfo
from frequenz.api.common.v1.pagination.pagination_params_pb2 import PaginationParams
```

src/frequenz/client/common/pagination/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+    """Parameters for paginating list requests."""
+
+    page_size: int | None = None
+    """The maximum number of results to be returned per request."""
+
+    page_token: str | None = None
+    """The token identifying a specific page of the list results."""
+
+    @classmethod
+    def from_proto(cls, pagination_params: PBPaginationParams) -> Self:
```
llucax:
```suggestion
    def from_proto(cls, pagination_params: PaginationParams) -> Self:
```

src/frequenz/client/common/pagination/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+            pagination_params: Params to convert.
+        Returns:
+            Params object corresponding to the protobuf message.
+        """
+        return cls(
+            page_size=pagination_params.page_size,
+            page_token=pagination_params.page_token,
+        )
+
+    def to_proto(self) -> PBPaginationParams:
```
llucax:
```suggestion
    def to_proto(self) -> PaginationParams:
```

src/frequenz/client/common/pagination/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+        """Convert a Params object to protobuf PaginationParams.
+
+        Returns:
+            Protobuf message corresponding to the Params object.
+        """
+        return PBPaginationParams(
```
llucax:
```suggestion
        return PaginationParams(
```

src/frequenz/client/common/pagination/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+    """Information about the pagination of a list request."""
+
+    total_items: int
+    """The total number of items that match the request."""
+
+    next_page_token: str | None = None
+    """The token identifying the next page of results."""
+
+    @classmethod
+    def from_proto(cls, pagination_info: PBPaginationInfo) -> Self:
```
llucax:
```suggestion
    def from_proto(cls, pagination_info: PaginationInfo) -> Self:
```

src/frequenz/client/common/pagination/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+            pagination_info: Info to convert.
+        Returns:
+            Info object corresponding to the protobuf message.
+        """
+        return cls(
+            total_items=pagination_info.total_items,
+            next_page_token=pagination_info.next_page_token,
+        )
+
+    def to_proto(self) -> PBPaginationInfo:
```
llucax:
```suggestion
    def to_proto(self) -> PaginationInfo:
```

src/frequenz/client/common/pagination/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+        """Convert a Info object to protobuf PBPaginationInfo.
+
+        Returns:
+            Protobuf message corresponding to the Info object.
+        """
+        return PBPaginationInfo(
```
llucax:
```suggestion
        return PaginationInfo(
```

llucax (commented):
If you still want to rename the imports for clarity, you can do so, but now it is not necessary any more because we are not using the same class names.

```py
from frequenz.api.common.v1.pagination.pagination_info_pb2 import PaginationInfo as PBPaginationInfo
from frequenz.api.common.v1.pagination.pagination_params_pb2 import PaginationParams as PBPaginationParams
```

llucax (commented):
This looks good to me, thanks for the patience!

There are a few small changes I would like to propose, but I will do it in a separate PR so we don't block this any longer, I think it is already in a good shape to be merged, and we can build on top of it.

I would strongly suggest to squash all the commits into one to avoid having 41 commit for such a small final change :smile: 

I will approve after it is squashed, let me know if you need any help with that!

Marenz (approved):

llucax (approved):

llucax:
This is still waiting for a GPG signature for the commit @flora-hofmann-frequenz

PR #21, opened 2024-03-05
Add common metric to test with reporting client mvp

flora-hofmann-frequenz:
This PR is aimed to work with the [mvp for the reporting client](https://github.com/frequenz-floss/frequenz-api-reporting/pull/20/files).

The code is taken from discussion heavy PR #8 (with PR #8 being turned into _draft_ for the time being), and is similar to component types PR #11.


src/frequenz/client/common/metric/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+from frequenz.api.common.v1.metrics.metric_sample_pb2 import Metric as PBMetric
+
+# pylint: enable=no-name-in-module
+
+
+class Metric(Enum):
+    """List of supported metrics."""
+
+    # Default value
+    UNSPECIFIED = PBMetric.METRIC_UNSPECIFIED
```
cwasicki, line 19: I think the doc for each field needs to be copied from the proto specs.

flora-hofmann-frequenz: Happy to add them!

src/frequenz/client/common/metric/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+    SENSOR_WIND_SPEED = PBMetric.METRIC_SENSOR_WIND_SPEED
+    SENSOR_WIND_DIRECTION = PBMetric.METRIC_SENSOR_WIND_DIRECTION
+    SENSOR_TEMPERATURE = PBMetric.METRIC_SENSOR_TEMPERATURE
+    SENSOR_RELATIVE_HUMIDITY = PBMetric.METRIC_SENSOR_RELATIVE_HUMIDITY
+    SENSOR_DEW_POINT = PBMetric.METRIC_SENSOR_DEW_POINT
+    SENSOR_AIR_PRESSURE = PBMetric.METRIC_SENSOR_AIR_PRESSURE
+    SENSOR_IRRADIANCE = PBMetric.METRIC_SENSOR_IRRADIANCE
+
+    @classmethod
+    def from_proto(cls, metric: PBMetric.ValueType) -> Metric:
```
llucax:
```suggestion
    def from_proto(cls, metric: PBMetric.ValueType) -> Self:
```

src/frequenz/client/common/metric/__init__.py
```diff
@@ -0,0 +1,141 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Module to define the metrics used with the common client."""
+
+from __future__ import annotations  # required for constructor type hinting
+
```
llucax: This is not needed if you use `Self`:

```suggestion
from typing import Self

```

llucax (commented):
LGTM after docs are added.

I think we should keep in mind that all of this could change in the future, but other than that, I would say let's go ahead. Also enums are very likely to stay as is.

src/frequenz/client/common/metric/__init__.py
flora-hofmann-frequenz: The docs are the same as in the [proto](https://github.com/frequenz-floss/frequenz-api-common/blob/2a28ec452861cde9afb520cb6cb67f8a30abf8fb/proto/frequenz/api/common/v1/metrics/metric_sample.proto#L65), I've checked again. Or did you mean something different?

cwasicki: Oh, sorry, I thought there was a docstring for each field. Looks good then.

cwasicki (dismissed):

src/frequenz/client/common/metric/__init__.py
flora-hofmann-frequenz: I actually get a `mypy` error from using `Self` here > not sure how to change that now?

flora-hofmann-frequenz: I get a mypy error if I use `Self` instead of `Metric`.

cwasicki (dismissed):

src/frequenz/client/common/metric/__init__.py
flora-hofmann-frequenz: Discussed with @Marenz and resolved.

Marenz (approved):

PR #31, opened 2024-05-16
Update dependencies in line with api-common release

flora-hofmann-frequenz:
New api-common [v0.6.1](https://github.com/frequenz-floss/frequenz-api-common/releases/tag/v0.6.1) has been released.


pyproject.toml
```diff
@@ -27,7 +27,7 @@ classifiers = [
 requires-python = ">= 3.11, < 4"
 dependencies = [
   "typing-extensions >= 4.5.0, < 5",
-  "frequenz-api-common >= 0.5.4, < 6",
+  "frequenz-api-common >= 0.6.1, < 7",
```
Marenz: I think we don't need to bump the min version, that way others can still use the older dependency until they need it

flora-hofmann-frequenz: Sounds good! I will change it accordingly.

llucax: No, sadly v0.6.0 introduced breaking changes without using a different *namespace*, so the dependency needs to be bumped because (if I understand correctly), this PR is actually updating to use the new names.

I guess since this repo doesn't change that much for now, we could keep a branch that's compatible with v0.5.x and one that's compatible with v0.6.x **if necessary**. But I guess it is only reporting using this repo for now so that's probably not necessary.

llucax (approved):
Approving because I think the bump to 0.6.0 is necessary, but please double check if this is correct.

llucax:
BTW, remember to close the milestone when you release :)

pyproject.toml
flora-hofmann-frequenz: Speaking for reporting, we would need v0.6.x for the new names (no need to have a branch that's compatible with v0.5.x in our case).

Marenz: Let's just do that on demand, no need to add extra work without cause

PR #39, opened 2024-08-01
Update metric info

flora-hofmann-frequenz:
Add information to `energy` metric to inform users about drawbacks of this metric.

### DIFF
--- a/src/frequenz/client/common/metric/__init__.py
+++ b/src/frequenz/client/common/metric/__init__.py
@@ -13,7 +13,18 @@ from frequenz.api.common.v1.metrics.metric_sample_pb2 import Metric as PBMetric
 
 
 class Metric(Enum):
-    """List of supported metrics."""
+    """List of supported metrics.
+
+    AC energy metrics information:
+    * This energy metric is reported directly from the component, and not a
+    result of aggregations in our systems. If a component does not have this
+    metric, this field cannot be populated.
+    * Components that provide energy metrics reset this metric from time to
+    time. This behaviour is specific to each component model. E.g., some
+    components reset it on UTC 00:00:00.
+    * This energy metric does not specify the timestamp since when the energy
+    was being accumulated, and therefore can be inconsistent.
+    """
 
     # Default value
     UNSPECIFIED = PBMetric.METRIC_UNSPECIFIED
@@ -57,7 +68,7 @@ class Metric(Enum):
     AC_POWER_FACTOR_PHASE_2 = PBMetric.METRIC_AC_POWER_FACTOR_PHASE_2
     AC_POWER_FACTOR_PHASE_3 = PBMetric.METRIC_AC_POWER_FACTOR_PHASE_3
 
-    # AC energy metrics
+    # AC energy metrics - Please be careful when using and check Enum docs
     AC_APPARENT_ENERGY = PBMetric.METRIC_AC_APPARENT_ENERGY
     AC_APPARENT_ENERGY_PHASE_1 = PBMetric.METRIC_AC_APPARENT_ENERGY_PHASE_1
     AC_APPARENT_ENERGY_PHASE_2 = PBMetric.METRIC_AC_APPARENT_ENERGY_PHASE_2
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,12 +6,24 @@
 
 ## Upgrading
 
+<<<<<<< HEAD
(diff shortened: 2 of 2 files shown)


flora-hofmann-frequenz:
@tiyash-basu-frequenz Can you pls check if I mentioned all drawbacks of the energy metric correctly?

src/frequenz/client/common/metric/__init__.py
```diff
@@ -57,7 +57,12 @@ class Metric(Enum):
     AC_POWER_FACTOR_PHASE_2 = PBMetric.METRIC_AC_POWER_FACTOR_PHASE_2
     AC_POWER_FACTOR_PHASE_3 = PBMetric.METRIC_AC_POWER_FACTOR_PHASE_3
 
-    # AC energy metrics
+    """AC energy metrics
+    Not all components will have energy metrics as components tent to reset
+    to avoid overflow. This depends on the component model.
+    Additionally, the energy metric is inconsistent as it does not provide
+    a time parameter.
```
tiyash-basu-frequenz: typo `tent` => `tend`

tiyash-basu-frequenz: There are 3 separate points here:

- This energy metric is reported directly from the component, and not a result of aggregations i our systems. If a component does not have this metric, this field cannot be populated.
- Components that provide energy metrics reset this metric from time to time. This behaviour is specific to each component model. E.g., some components reset it on UTC 00:00:00.
- This energy metric does not specify the timestamp since when the energy was being accumulated, and therefore can be inconsistent.

In the original text, it reads like points 1 and 2 were linked. But they are not.

src/frequenz/client/common/metric/__init__.py
```diff
@@ -57,7 +57,16 @@ class Metric(Enum):
...
+    """AC energy metrics
+    * This energy metric is reported directly from the component, and not a
+    result of aggregations in our systems. If a component does not have this
+    metric, this field cannot be populated.
+    * Components that provide energy metrics reset this metric from time to
+    time. This behaviour is specific to each component model. E.g., some
+    components reset it on UTC 00:00:00.
+    * This energy metric does not specify the timestamp since when the energy
+    was being accumulated, and therefore can be inconsistent.
+    """
```
llucax: As it is, this docstring is documenting the varaible `AC_POWER_FACTOR_PHASE_3` (docstrings always document the symbol above, not below it).

Since we have multiple energy metrics, I think it should be best to document this in the `Enum` class, and in the metric just say you have to be careful and to read the `Enum` docs, so we don't have a big explanation duplicated in many metrics.

You can always check how generated docs look like by doing `pip install -e .[dev]; mkdocs serve`

src/frequenz/client/common/metric/__init__.py
```diff
@@ -13,7 +13,18 @@
...
+
+    AC energy metrics information:
+    * This energy metric is reported directly from the component, and not a
+    result of aggregations in our systems. If a component does not have this
+    metric, this field cannot be populated.
+    * Components that provide energy metrics reset this metric from time to
+    time. This behaviour is specific to each component model. E.g., some
+    components reset it on UTC 00:00:00.
+    * This energy metric does not specify the timestamp since when the energy
+    was being accumulated, and therefore can be inconsistent.
```
Marenz, line 26: Feels almost like "USE AT YOUR OWN RISK" disclaimer XD

Marenz:
@flora-hofmann-frequenz you committed a conflict and then a merge commit that resolves it, I think a rebase might be required :)

src/frequenz/client/common/metric/__init__.py
tiyash-basu-frequenz: In a sense, yes. We decided to be transparent to the client, and expose whatever metric the components provide. We can sanitise a few, but not all, e.g., energy.

I guess we will get more user feedback for this field in the future ..

llucax (approved):

llucax:
> @flora-hofmann-frequenz you committed a conflict and then a merge commit that resolves it, I think a rebase might be required :)

Damn, I missed this comment, just followed the notification and saw the full final diff :facepalm:

flora-hofmann-frequenz:
> > @flora-hofmann-frequenz you committed a conflict and then a merge commit that resolves it, I think a rebase might be required :)
> 
> Damn, I missed this comment, just followed the notification and saw the full final diff 🤦

...sorry I also missed it in between all Monday meetings. Can we leave it as is now?!

PR #44, opened 2024-11-14
Add component states and error codes

cyiallou:
Add component state codes and error codes similar to #11

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -10,7 +10,8 @@ Update of the Pagination `Params` data class.
 
 ## New Features
 
-* Additional information for energy metric
+* Additional information for energy metric.
+* Add component state codes.
 
 ## Bug Fixes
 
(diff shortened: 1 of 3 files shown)


cyiallou:
Rebased to latest main branch.

llucax (approved):
LGTM.

I have done a more high-level abstraction of all microgrid components in the upgrade of the microgrid client to the microgrid API v0.17 (which also finally switches to the v1 in common), so this will probably change in a near future.

We need to talk to see if this abstraction I wrote makes sense for other services too, but in case you want to give some early feedback, it is part of this PR:

https://github.com/frequenz-floss/frequenz-client-microgrid-python/pull/94/files#diff-920815d585fffddae9f723936d74fbddddf8176ee16d4eb2a557e634d22e1d12

My idea is to basically move everything in `src/frequenz/client/microgrid/component` and `metrics`, as well as the top-level modules related to `_id.py`, `_location.py`, `_microgrid_info.py`, `_lifetime.py` and `_delivery_area.py` to this repo.

PR #72, opened 2025-05-13
Add conftest to allow for wider range of Protobuf version exceptions

flora-hofmann-frequenz:


tests/conftest.py
```diff
@@ -0,0 +1,12 @@
+# License: MIT
+# Copyright © 2025 Frequenz Energy-as-a-Service GmbH
+
+"""Show Protobuf version warning during tests, but don't treat as error."""
+
+import warnings
+
+warnings.filterwarnings(
+    "default",
```
llucax: I would use once here, just in case, we only need to see the warning once.

```suggestion
    "once",
```

llucax (approved):

PR #77, opened 2025-06-06
Add microgrid-related ID types

llucax:


llucax:
I created a v0.3.x branch for this because v0.x.x already depends on `api-common` v0.x.x (to be v0.7.0). Eventually this will be merged to v0.x.x too.

llucax:
Ready for review, but converting to draft until core 1.0 is released.

ela-kotulska-frequenz:
Does it mean that now `ComponentId` and `MicrogridId` will be defined both here and in `client.microgrid` ?

llucax:
> Does it mean that now `ComponentId` and `MicrogridId` will be defined both here and in `client.microgrid` ?

No, `client.microgrid` will import it from here, and all other APIs needing them (like dispatch) should import from here too.

llucax:
This is ready for review now.

src/frequenz/client/common/microgrid/components/__init__.py
```diff
@@ -16,10 +18,16 @@
...
+from frequenz.core.id import BaseId
 
 # pylint: enable=no-name-in-module
 
 
+@final
+class ComponentId(BaseId, str_prefix="CID"):
+    """A unique identifier for a microgrid component."""
+
+
```
ela-kotulska-frequenz, line 30: But definition ComponentId in client.microgrid  is different. The original version checks if component_id != 0.
Why Is this version different?

llucax: IIRC that was a bogus limitation. @tiyash-basu-frequenz ?

ela-kotulska-frequenz: I remember it was there because grid has always id=0.
But I don't know if it is still true.

shsms: The component IDs are generated automatically now.  So 0 is allowed and the IDs are unique across enterprises

ela-kotulska-frequenz (approved):

PR #85, opened 2025-07-01
Merge tag 'v0.3.3' into v0.x.x

llucax:
- A new module `frequenz.client.common.enum_proto` has been added, which provides a generic `enum_from_proto()` function to convert protobuf enums to Python enums.
- Documentation for some `frequenz.client.common.microgrid.electrical_components.ElectricalComponentCategory` values was improved.
- The metrics and components enums `.from_proto()` are deprecated, please use the new `enum_from_proto()` instead.
- Classes to represent microgrid-related ID were added (`MicrogridId`, `EnterpriseId`, `ComponentId`, and `SensorId`).

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,16 +2,18 @@
 
 ## Summary
 
-<!-- Here goes a general summary of what this release is about -->
+This release replaces the failed v0.3.1 release.
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+- The `typing-extensions` dependency minimum version was bumped to 4.13 to support Python 3.12.
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+- New `BaseId` class to create unique IDs for entities in the system.
+- New ID classes for microgrid-related entities:
 
-## Bug Fixes
-
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+   * `EnterpriseId`
+   * `MicrogridId`
+   * `ComponentId`
+   * `SensorId`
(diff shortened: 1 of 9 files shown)


llucax:
Question: The ID type prefix for Electrical components before they were explicitly named "electrical" was `CID`. We use this prefix everywhere, so I left it as `CID`, even when now just `CID` might be ambiguous as we have non-electrical component IDs too. I guess when we add *Communication Component IDs* we will call them `CCID` and we keep the unique prefixes, but now maybe it would be too easy oversee the difference between `CCID` and `CID` so they might get confused.

So I wonder if you think we should bite the bullet and just go with `ECID` for *electrical component IDs` and slowly update our use of `CID`, or keep `CID` for historical reasons?

@cwasicki @tiyash-basu-frequenz @shsms 

(for more context, the prefix is used when converting a ID object to a string, for example when logging an ID)

Marenz (approved):
Not quite sure how to review this :sweat_smile:

cwasicki:
> (for more context, the prefix is used when converting a ID object to a string, for example when logging an ID)

So it mainly affects logging? Don't have objections against neither of the two options.

llucax:
Yeah, it is never easy to review merges, but I would say this:

* https://github.com/frequenz-floss/frequenz-client-common-python/pull/85/commits/7098a3f61aba53fd563e4913fc270e17f621b696?diff=split

llucax:
> So it mainly affects logging? Don't have objections against neither of the two options.

Yeah, logging and how people refer to an electrical component ID, as if they don't match it can get pretty confusing.

llucax:
Actually now that I think about it, I guess `CID` is not **that** widely used, I think `MID` is much more widely used (but there is no need to change that one), so maybe going for `ECID` is not that big of a change.

Marenz:
Closing in favor of https://github.com/frequenz-floss/frequenz-client-common-python/pull/97

PR #96, opened 2025-08-07
feat: Update frequenz-api-common to v0.8 and adapt to API changes

Marenz:
This PR introduces a new module v1alpha8 that provides all the updates/changes that the common-api introduces. It also reverts the renaming of Components to ElectricalComponents to stay backwards compatible (the rename was not yet released)


cwasicki (dismissed):

src/frequenz/client/common/v1alpha8/pagination/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+)
+from frequenz.api.common.v1alpha8.pagination.pagination_params_pb2 import (
+    PaginationParams,
+)
+
+# pylint: enable=no-name-in-module
+
+
+@dataclass(frozen=True, kw_only=True)
+class Params:
```
shsms: better keep the original name, no?  This could lead to ambiguous code otherwise.

```suggestion
class PaginationParams:

File: src/frequenz/client/common/v1alpha8/pagination/__init__.py [hunk 2]
```diff
@@ -0,0 +1,91 @@
...
+            Protobuf message corresponding to the Params object.
+        """
+        return PaginationParams(
+            page_size=self.page_size,
+            page_token=self.page_token,
+        )
+
+
+@dataclass(frozen=True, kw_only=True)
+class Info:
```
shsms:
```suggestion
class PaginationInfo:

src/frequenz/client/common/v1alpha8/pagination/__init__.py
Marenz: I copied it from v1 (non-versioned in this repo) like that. But true, we can use this opportunity to rename it in alpha8

cwasicki (approved):

llucax:
I'm not convinced we should just create a new namespace here as we do with protobuf. Unlike protobuf messages, which are wire-compatible, adding more classes to python could be problematic (mainly because we use type hints, if we didn't then things would work like with protobuf). If someone imports from one namespace and then try to pass the instance to a function that expects an instance from a different namespace, then things will break.

In particular classes inheriting from `BaseId` **WILL BREAK**, as they share the same prefix, so we'll get an exception as soon as someone tries to import an `Id` class from 2 namespaces.

We need to give more thought to this.

llucax:
* https://github.com/frequenz-floss/frequenz-client-common-python/issues/100

PR #97, opened 2025-08-11
Merge v0.3.x into v0.x.x

Marenz:
This merges all v0.3.x changes into the v0.x.x branch.

I assume we had an extra v0.3.x branch due to the rename changes of Components -> ElectricalComponents.
Since then we decided to handle those API updates using extra paths like `.v1alpha8`, so we can add back the changes of v0.3.x and stay API compatible with our dependees.

- A new module `frequenz.client.common.enum_proto` has been added, which provides a generic `enum_from_proto()` function to convert protobuf enums to Python enums.
- Documentation for some `frequenz.client.common.microgrid.electrical_components.ElectricalComponentCategory` values was improved.
- The metrics and components enums `.from_proto()` are deprecated, please use the new `enum_from_proto()` instead.
- Classes to represent microgrid-related ID were added (`MicrogridId`, `EnterpriseId`, `ComponentId`, and `SensorId`).

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,16 +2,18 @@
 
 ## Summary
 
-<!-- Here goes a general summary of what this release is about -->
+This release replaces the failed v0.3.1 release.
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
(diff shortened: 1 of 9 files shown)


Marenz:
I just discovered that this is a dup of #85.  As it resolved the conflicts and is effectively the same and #85 is already approved, I will just go ahead and merge this if noone is gonna approve soon :)

PR #103, opened 2025-08-18
Add v1alpha8 variant of pagination structs

Marenz:
- **Add v1alpha8 variant of pagination structs**

### DIFF
--- a/src/frequenz/client/common/pagination/__init__.py
+++ b/src/frequenz/client/common/pagination/__init__.py
@@ -124,9 +124,17 @@ class PaginationInfo:
         Returns:
             Info object corresponding to the protobuf message.
         """
+        # We check for truthiness here to handle both cases where the token is
+        # not set (defaults to "") or is explicitly set to "". In both
+        # situations, we want to return `None`. Using `HasField("next_page_token")`
+        # would not handle the case where the token is explicitly set to "".
         return cls(
             total_items=pagination_info.total_items,
-            next_page_token=pagination_info.next_page_token,
+            next_page_token=(
+                pagination_info.next_page_token
+                if pagination_info.next_page_token
+                else None
+            ),
         )
 
     def to_proto_v1alpha8(self) -> PBPaginationInfoAlpha8:
--- a/src/frequenz/client/common/pagination/__init__.py
+++ b/src/frequenz/client/common/pagination/__init__.py
@@ -9,12 +9,25 @@ from dataclasses import dataclass
 from typing import Self
 
 # pylint: disable=no-name-in-module
-from frequenz.api.common.v1.pagination.pagination_info_pb2 import PaginationInfo
-from frequenz.api.common.v1.pagination.pagination_params_pb2 import PaginationParams
+from frequenz.api.common.v1.pagination.pagination_info_pb2 import (
+    PaginationInfo as PBPaginationInfo,
+)
+from frequenz.api.common.v1.pagination.pagination_params_pb2 import (
+    PaginationParams as PBPaginationParams,
+)
+from frequenz.api.common.v1alpha8.pagination.pagination_info_pb2 import (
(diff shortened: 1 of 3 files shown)


src/frequenz/client/common/pagination/__init__.py
```diff
@@ -62,7 +80,79 @@ class Info:
...
+    """Parameters for paginating list requests."""
+
+    page_size: int
+    """The maximum number of results to be returned per request."""
+
+    page_token: str
+    """The token identifying a specific page of the list results."""
+
+    @classmethod
+    def from_proto(cls, pagination_params: PBPaginationParamsAlpha8) -> Self:
```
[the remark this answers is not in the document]
llucax: I partially agree with copilot.

To be able to really deprecate `Params`, we need to be able to load from the old API common menssage too, so this method should accept `PBPaginationParams | PBPaginationParamsAlpha8`.

src/frequenz/client/common/pagination/__init__.py
```diff
@@ -75,13 +92,85 @@ def from_proto(cls, pagination_info: PaginationInfo) -> Self:
...
+    def to_proto(self) -> PBPaginationParamsAlpha8:
+        """Convert a Params object to protobuf PaginationParams.
+
+        Returns:
+            Protobuf message corresponding to the Params object.
+        """
+        return PBPaginationParamsAlpha8(
+            page_size=self.page_size,
+            page_token=self.page_token,
+        )
```
llucax: And we should probably keep the old `to_proto()` here and add a `to_proto_v1alpha8()`.

src/frequenz/client/common/pagination/__init__.py
```diff
@@ -62,7 +80,79 @@ class Info:
...
+    """Information about the pagination of a list request."""
+
+    total_items: int
+    """The total number of items that match the request."""
+
+    next_page_token: str | None = None
+    """The token identifying the next page of results."""
+
+    @classmethod
+    def from_proto(cls, pagination_info: PBPaginationInfoAlpha8) -> Self:
```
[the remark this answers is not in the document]
llucax: Same comments as for `PaginationParams`.

src/frequenz/client/common/pagination/__init__.py
```diff
@@ -75,13 +92,85 @@ def from_proto(cls, pagination_info: PaginationInfo) -> Self:
...
+        """Convert a protobuf PBPaginationInfo to Info object.
+
+        Args:
+            pagination_info: Info to convert.
+        Returns:
+            Info object corresponding to the protobuf message.
+        """
+        return cls(
+            total_items=pagination_info.total_items,
+            next_page_token=pagination_info.next_page_token,
```
llucax: This also needs fixing (we can fix it only here if we deprecate the old one):
```suggestion
            next_page_token=pagination_info.next_page_token if pagination_info.HasField("next_page_token") else None,
```

Otherwise it will be assigned to `""` when the field is missing.

llucax (commented):
So we deprecate `Params` and remove `PaginationParams` too, right?

tests/test_pagination.py
```diff
@@ -0,0 +1,64 @@
...
+    assert info.next_page_token == "token"
+
+
+def test_deprecated_info_to_proto() -> None:
+    """Test the deprecated Info to_proto method."""
+    with pytest.deprecated_call():
+        info = Info(total_items=100, next_page_token="token")
+    proto = info.to_proto()
+    assert proto.total_items == 100
+    assert proto.next_page_token == "token"
```
llucax, line 96: Also add a test here for `i = PaginationInfo.from_proto(PBPaginationInfo(total_items=100)); assert i.next_page_token is None`.

llucax (approved):

PR #140, opened 2026-01-23
Add metric from/to protobuf v1alpha8 conversion

llucax:
To make sure we can decouple the clients from the underlying protocol, we need to make sure we always use explicit conversion between Python enums and protobuf enums. This is especially important to be able to support multiple versions of a protobuf message, for example an upcoming v1alpha9 Metric enum, so downstream project can migrate to new versions in a backwards-compatible way.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,13 +4,17 @@
 
 <!-- Here goes a general summary of what this release is about -->
 
-## Upgrading
+## Deprecation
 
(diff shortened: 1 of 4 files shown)


llucax:
Part of #141.

Marenz (dismissed):

llucax:
Needed to reformat with black due to a new black version after rebasing. Needs a new approval. @Marenz :pray:

Marenz (approved):

PR #181, opened 2026-06-01
Add conversion for the remaining `v1alpha8` enums

llucax:
This PR is a followup on #140 and #142. 

It uses plan `int`s as enum values, deprecates implicit conversions, and add explicit conversion functions for the rest of the `v1alpha8` enums: `MetricConnectionCategory`, `ElectricalComponent*` and `Event`.

It also adds new enum values up to api-common 0.8.4 and an utility test class to easily generate the parity tests for all enums.

Part of #141.

PR #182, opened 2026-06-01
Add a new `grid` package with delivery area definitions

llucax:
Added a new `frequenz.client.common.grid` package with `DeliveryArea` and `EnergyMarketCodeType` definitions for representing energy delivery areas, including `v1alpha8` conversion functions in `frequenz.client.common.grid.proto.v1alpha8`.

The code was copied from the `frequenz-client-microgrid` v0.18.3 and adapted to this repository.


llucax:
Draft because it is based in #181.

src/frequenz/client/common/grid/proto/v1alpha8/_delivery_area.py
```diff
@@ -0,0 +1,73 @@
...
+    Args:
+        code_type: An EnergyMarketCodeType enum member.
+
+    Returns:
+        The corresponding protobuf EnergyMarketCodeType enum value.
+    """
+    return delivery_area_pb2.EnergyMarketCodeType.ValueType(code_type.value)
+
+
+def delivery_area_from_proto(message: delivery_area_pb2.DeliveryArea) -> DeliveryArea:
```
daniel-zullo-frequenz, line 45: delivery_area_to_proto seems missing, is it needed?

daniel-zullo-frequenz (commented):
I have only a question, LGTM

src/frequenz/client/common/grid/proto/v1alpha8/_delivery_area.py
llucax: We can add it when anyone needs it, at least on the microgrid API client and assets API client I think we only need to parse it, it is created only on the service side which uses rust, so we don't need a Python function to convert **to** proto.

@cwasicki do you have any use case for `to_proto()`?

Marenz (approved):

src/frequenz/client/common/grid/proto/v1alpha8/_delivery_area.py
cwasicki: Wouldn't it be useful if we want to create a filter on delivery area, e.g. here https://github.com/frequenz-floss/frequenz-api-assets/pull/99/changes#diff-f8b6949122d5be6141be00a4dfe98e6b508f5d8512059edcd335875aa706e209R509

llucax: I'm not sure, those filters are complicated. For example delivery area can only be set if gridpool ID is also set  Maybe we can have a safer and simpler interface in Python where we take directly the delivery area fields in the call instead of a delivery area object. I would wait a bit to see if we really need it.

Issue #141, opened 2026-01-23
Allow for smooth upgrading between underlaying protocol versions


llucax:
### What's needed?

We need to be able to upgrade to newer protobuf versions without breaking all user code.

### Proposed solution

In general:

* Always require using conversion functions between the provided Python wrappers and the protobuf messages, even for enums.
* Group conversion functions by protobuf version to ensure only the necessary protobuf files are imported to avoid bloat.
* Be more explicit about for which protobuf version conversion functions are.

In particular:

* Stop using protobuf values as enum values in wrappers
  * Using a protobuf values ties the wrapper to a particular protobuf version
  * It forces to import a particular version of protobuf messages even if we want to use a different version, causing import bloat
  * Allows to accidentally pass values to protobuf functions
* Use simple `int` values as enum values instead
  * Consider enum values opaque, using the values besides printing them for debugging should be forbidden
  * This allows us to still maintain a 1-1 mapping of values with the latest protobuf version for improved debuggability in most cases (people using the latest version)
* Add a `.proto.<version>` (for example `frequenz.client.common.metrics.proto.v1alpha8`) package containing the conversion functions for each supported protobuf version

The new conversion functions should be added in v0.3.x, as well as doing as much deprecation as possible to guide users towards an upgrade to v0.4.0 that is as smooth as possible.

Enum members should keep their current values in v0.3.x, and only be updated to `int`s in v0.4.0 to ensure v0.3.x stays backwards compatible, but allowing users to migrate to use conversion functions at their own pace. If they do so before v0.4.0, the upgrade to v0.4.0 should not even be a breaking change.

### Additional context

This approach gives full flexibility to manage protocol-level breaking changes in a much more graceful way, allowing us to provide backwards-compatible upgrade paths via deprecations.

llucax:
I submitted 2 pull requests as a proof of concept for this:

* Backwards-compatible changes, just adding the conversion functions, that for now are trivial, and testing the protobuf and Python enum match. Users can already upgrade to use the new conversion functions.

    #140

[...]

PR #193, opened 2026-06-08
Reorganise subpackages to mirror the `proto/v1alpha8` layout

llucax:
This is a followup on #181 and #182.

It moves the per-domain types and conversions into their own modules so the source layout mirrors the established `proto/v1alpha8/` structure used by the other domains:

- `Event`, `SensorId`, the microgrid IDs, `ElectricalComponentId`, the electrical component enums and `PaginationInfo` are split into their own (private) modules.
- The `metrics` proto conversions move from `frequenz.client.common.metrics.proto` to `frequenz.client.common.metrics.proto.v1alpha8`, keeping the old import path as a deprecated shim.
- The tests are reorganised to mirror the new source layout.

It also adds explicit `v1alpha8` conversion functions for `PaginationInfo` (`pagination_info_from_proto`, `pagination_info_to_proto`) and deprecates the old `PaginationInfo.from_proto`/`to_proto`/`to_proto_v1alpha8` methods in favour of them.

Part of #141.


tests/metrics/proto/v1alpha8/__init__.py
```diff
@@ -0,0 +1,4 @@
+# License: MIT
+# Copyright © 2026 Frequenz Energy-as-a-Service GmbH
+
+"""Tests for the grid package."""
```
Marenz: Copy-paste nit: docstring says "Tests for the grid package" — should be something like "Tests for the metrics proto v1alpha8 package".

src/frequenz/client/common/pagination/_pagination_info.py
```diff
@@ -0,0 +1,94 @@
...
+        # pylint: disable-next=import-outside-toplevel,cyclic-import
+        from frequenz.client.common.pagination.proto.v1alpha8 import (
+            pagination_info_to_proto,
+        )
+
+        return pagination_info_to_proto(self)
+
+    @deprecated(
+        "frequenz.client.common.pagination.PaginationInfo.to_proto is deprecated. "
+        "Use frequenz.client.common.pagination.proto.v1alpha8."
```
Marenz, line 82: The deprecation message here points users to `pagination_info_to_proto` (which produces a v1alpha8 `PaginationInfoPb`), but this `to_proto()` method returns a v1 `PBPaginationInfo`. Users who actually need v1 output would silently get a different proto version after migrating. Might be worth either:
- noting the version difference in the deprecation message, or
- adding a separate v1 converter, or
- having `to_proto()` delegate to the v1alpha8 converter too (if they're wire-compatible).

src/frequenz/client/common/microgrid/electrical_components/_category.py
```diff
@@ -0,0 +1,115 @@
...
+        Args:
+            component_category: protobuf enum to convert
+
+        Returns:
+            Enum value corresponding to the protobuf message.
+        """
+        if not any(t.value == component_category for t in ElectricalComponentCategory):
+            return ElectricalComponentCategory.UNSPECIFIED
+        return cls(component_category)
+
```
Marenz, line 108: `to_proto()` is not deprecated here, while `from_proto()` is. `PaginationInfo` deprecates all its proto methods. Is the asymmetry intentional? If these `to_proto()` methods are also going away, might be good to deprecate them now for consistency.

Marenz (commented):
Nice refactoring — clean split, good backward-compat shims. A few small things inline.

src/frequenz/client/common/pagination/_pagination_info.py
llucax: Yeah, this on purpose, `v1` is already retired, so nobody should be using `v1` at the moment. I was planning to remove all `v1` stuff in v0.4.0.

src/frequenz/client/common/microgrid/electrical_components/_category.py
llucax: No, probably it was an oversight, since we don't have a replacement for `to_proto()` for `v1`, but we should deprecate it anyway and point to the `xxx_to_proto_v1alpha8()`.

llucax: Fixed in a new commit.

tests/metrics/proto/v1alpha8/__init__.py
llucax: Fixed the original commit.

llucax:
Enabled auto-merge.

Marenz (approved):

Issue #203, opened 2026-06-11
Import wrappers from the Microgrid and Assets API


llucax:
### What's needed?

We need to share the common wrappers between the microgrid and assets API, so the clients are interoperable.

### Proposed solution

Move the wrappers to this repo, modernize and generalize them, and adapt them to the common client repo layout.

PR #204, opened 2026-06-11
Add `ElectricalComponent` and problematic component classes

llucax:
Import `Component` abstract base from the microgrid client, rename it to `ElectricalComponent`.

Also import the `ProblematicComponent` abstract base class and the classes that inherit from them. This classes will be renamed in a follow-up PR.

Adapt the modules and tests to the common client layout and enable [mkdocstrings relative cross-referencing](https://mkdocstrings.github.io/python/usage/configuration/docstrings/?h=relative#relative_crossrefs).

Part of #203.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -108,6 +108,7 @@ plugins:
             docstring_section_style: spacy
             inherited_members: true
             merge_init_into_class: false
+            relative_crossrefs: true
             separate_signature: true
             show_category_heading: true
             show_root_heading: true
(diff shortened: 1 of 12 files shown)


llucax:
Skipping release notes, this will be part of a series of PRs adding all electrical components, and I will update the release notes at the end.

llucax:
It seems I need to do a bigger PR as this already refers (in the docs) to other components, and it doesn't make sense to *worsen* the docs to remove those references in this PR just to add them back in the next.

llucax:
OK, I added the *problematic* class hierarchy so tests pass. I also enabled relative cross-refs for mkdocstrings so cross-referencing becomes manageable.

llucax:
Ready for a review.

tiyash-basu-frequenz (approved):

PR #205, opened 2026-06-11
Add `Battery` wrapper classes

llucax:
Import the battery component module from the microgrid client with all its sub-classes and updates it to the client layout.

Part of #203.

### DIFF
--- a/src/frequenz/client/common/microgrid/electrical_components/_battery.py
+++ b/src/frequenz/client/common/microgrid/electrical_components/_battery.py
@@ -1,0 +1,131 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Battery component."""
+
+import dataclasses
+import enum
+from typing import Any, Literal, Self, TypeAlias
+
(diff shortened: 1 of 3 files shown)


src/frequenz/client/common/microgrid/electrical_components/_battery.py
```diff
@@ -0,0 +1,134 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
```
tiyash-basu-frequenz, line 2: year should be 2026, no?

Marenz: I nominate Luca for...

llucax: No, the file was created in 2024, we are just moving it, but we hold the copyright since 2024.

llucax: The idea to import verbatim was in part to make this obvious :laughing:

Marenz (approved):

PR #208, opened 2026-06-12
Add simple electrical component wrappers

llucax:
Import simple leaf component modules from the microgrid client:

- `Chp`
- `Converter`
- `CryptoMiner`
- `Electrolyzer`
- `Hvac`
- `Meter`
- `Precharger`
- `Relay`
- `SteamBoiler`
- `WindTurbine`

These components don't have any extra fields or methods, nor a sub-classes hierarchy, so we import all in one go.

We also adapt the docstrings, modules and tests to the common client layout.

In the future it might be worth having a reusable/parametrized test instead of having the same code copied over and over, but for now we are keeping the tests as they were in the microgrid client.

Part of #203.

### DIFF
--- a/src/frequenz/client/common/microgrid/electrical_components/__init__.py
+++ b/src/frequenz/client/common/microgrid/electrical_components/__init__.py
@@ -13,8 +13,12 @@ from ._battery import (
     UnspecifiedBattery,
 )
 from ._category import ElectricalComponentCategory
+from ._chp import Chp
+from ._converter import Converter
+from ._crypto_miner import CryptoMiner
 from ._diagnostic_code import ElectricalComponentDiagnosticCode
 from ._electrical_component import ElectricalComponent
+from ._electrolyzer import Electrolyzer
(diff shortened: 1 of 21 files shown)


tiyash-basu-frequenz (approved):
Approved.

But I wonder is this is a maintainable approach - this means we have to maintain all these wrappers, and add new ones when we add more component categories. Is there a reason behind this approach in this repo? Perhaps we can consider having a unified wrapper?

llucax:
Sure we can talk about it. I don't think maintenance is an issue though, initial creation took some time, yes, but once it is created, adding a component is just one small prompt away, and the approach is forward-compatible

A non-wrapped new component category doesn't mean you can't use it, you'll get an `UnrecognizedElectricalComponent` and you can use the raw component category as an `int` if you know how to interpret it. The idea is that in the happy path, the resulting code is very pythonic, and you can use `isinstance()` or `match` for example to check which type of component an object is.

PR #211, opened 2026-06-15
Add electrical component type aliases

llucax:
Import component type aliases from the microgrid client and adapt them to the common client layout.

Part of #203.

### DIFF
--- a/src/frequenz/client/common/microgrid/electrical_components/_types.py
+++ b/src/frequenz/client/common/microgrid/electrical_components/_types.py
@@ -1,0 +1,67 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""All known component types."""
+
+from typing import TypeAlias
+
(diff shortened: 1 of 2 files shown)


src/frequenz/client/common/microgrid/electrical_components/_types.py
```diff
@@ -0,0 +1,67 @@
...
+    | UnrecognizedEvCharger
+    | UnrecognizedInverter
+)
+
+ProblematicComponentTypes: TypeAlias = (
+    MismatchedCategoryComponent | UnrecognizedComponentTypes | UnspecifiedComponentTypes
+)
+"""All possible component types that has a problem."""
+
+ComponentTypes: TypeAlias = (
```
llucax, line 49: This actually should have been renamed `ElectricalComponentTypes` but I will do that in a follow PR, as I already have a few upcoming PRs ready to be sent sequentially, and I also noticed I didn't rename `ProblematicComponent` and subclasses to add the electrical prefix, so I will do all the proper renames in a separate PR.

daniel-zullo-frequenz (approved):

PR #216, opened 2026-06-16
Sync `ElectricalComponent` with the protobuf message

llucax:
The protobuf have a few new and deprecated fields and one existing field had a name that doesn't match the protobuf name, which could lead to confusion.

Since this code is new, we completely remove the deprecated fields.

Part of #203.

### DIFF
--- a/src/frequenz/client/common/microgrid/electrical_components/__init__.py
+++ b/src/frequenz/client/common/microgrid/electrical_components/__init__.py
@@ -45,6 +45,7 @@ from ._inverter import (
     UnspecifiedInverter,
 )
 from ._meter import Meter
+from ._operational_mode import ElectricalComponentOperationalMode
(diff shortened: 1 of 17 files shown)


llucax:
@tiyash-basu-frequenz I guess the new "model" field should be working by now and there is no need to keep the old "manufacturer" and "model_name" fields, right? Or should I leave them as deprecated as they are in the protobuf?

tiyash-basu-frequenz:
> @tiyash-basu-frequenz I guess the new "model" field should be working by now and there is no need to keep the old "manufacturer" and "model_name" fields, right? Or should I leave them as deprecated as they are in the protobuf?

The new field is indeed working now.

llucax:
Great. Rebased with resolved conflicts.

tiyash-basu-frequenz (approved):

PR #220, opened 2026-06-19
Remove unnecessary enums

### LINKED ISSUE #223: Remove all `UNSPECIFIED` enum values and low-level enums
### What's needed?

For the following enums:

- [ ] `Metric` (#221)
- [ ] `Event` (#225)

- [ ] `BatteryType` (#222)
- [ ] `EvChargerType` (#222)
- [ ] `InverterType` (#222)

llucax:
This is the fist PR in what it will very likely be a series of 3 PRs to remove unnecessary enums and eventually all `UNSPECIFIED` enum values.

Having `UNSPECIFIED` enum values is a common anti-pattern, and was only used so far because we started writing wrappers in a very 1-1 fashion. But Python abstractions should be more Pythonic and safer.

The broad plan is to better abstract whole enums where possible, and to remove `UNSPECIFIED` values for the rest, preferring using `None` (or omitting the value completely) for the unspecified case.

This PR mainly removes the `MicrogridStatus` and `ElectricalComponentOperationalMode` enums, using getters that can raise a new `UnspecifiedValueError` exception instead. Leaving the enums as low-level artifacts in the protobuf conversion functions only.

It also introduces and related safety feature, forbidding by default the construction of `Microgrid` and `ElectricalComponent` objects directly, as on the Python side they should only be retrieved from API services. This avoid users accidentally creating objects with *unspecified* values, another abstraction leak we want to avoid.

Part of #223.


llucax:
This might be a controversial change, but I think it is the right approach towards a more pythonic and save interface.

The plan for the following PRs is, next do the same for stuff that exists in v0.4.0, that should get a deprecation path (I will probably backport these deprecations to v0.3.x and consider the v0.4.0 release a glitch and go for v0.5.0 for the next release). This includes removing `UNSPECIFIED` for enums that we definitely need to keep, like `Metric` and most telemetry stuff (diagnostics, etc.).

Then the most disruptive and potentially controversial change: **completely remove the `ElectricalComponentCategory` and `XxxType` (battery, inverter, etc.) enums**. Where we use the enum now, we can pass the class instead, and this repo will provide a convenience function to convert from class type to protobuf enum for client implementors to use (for example for dispatch targets, or microgrid/assets filters when getting components).

Please comment if you have any concerns.

llucax:
I submitted a *phase 2* PR based on this one, so you can get a clearer picture of where this is going:

* #221

llucax:
If this approach is accepted, I will update the changes in this PR to also use `bool | int` as the low-level storage for the enum-based attributes (for example `_accepts_control: bool | int`), so we can raise `UnspecifiedValueError` and `UnrecognizedValueError` if a new enum member is ever added to these enums (which I guess it would be extremely rare, but better safe than sorry).

Marenz (approved):
:robot: LGTM. The enum removal and the `is_active()`/`provides_telemetry()`/`accepts_control()` accessors are a clear improvement, and the proto layer keeps the raw enum knowledge nicely contained. Tests are thorough.

One non-blocking thought on the `_allow_construction` gate: it works, but it's a soft guard — a leading-underscore kwarg users can still pass to bypass it, and it adds a second construction check alongside the existing `__new__` abstractness guard, with subclasses needing to remember `super().__post_init__()` (e.g. `GridConnectionPoint`) or they silently skip the gate. Not asking to change it here, but might be worth revisiting if it proliferates across the next PRs in the series.

llucax:
> One non-blocking thought on the _allow_construction gate: it works, but it's a soft guard — a leading-underscore kwarg users can still pass to bypass it,

Yeah, this is OK, we just want to avoid people accidentally constructing the class when they normally shouldn't.

> and it adds a second construction check alongside the existing __new__ abstractness guard, with subclasses needing to remember super().__post_init__() (e.g. GridConnectionPoint) or they silently skip the gate. Not asking to change it here, but might be worth revisiting if it proliferates across the next PRs in the series.

Yeah, this is a good point. I will note it down to research in the future if there is an alternative that only requires adding hacks to `__new__` as we already one there. For now I consider it low-prio as adding new component categories is rare (specially the complex ones that need to remember to define a `__post_init__`).

llucax:
If some human can take some time to read this proposal (3rd and final PR in the series coming soon) and comment on the proposal itself, not just the code that implements it, it would be appreciated.

If I don't get any comments, I will move forward with it.

cwasicki:
Without reviewing this in detail, the proposal looks clean to me.

tests/microgrid/electrical_components/test_simple_components.py
```diff
@@ -57,17 +56,24 @@ def microgrid_id() -> MicrogridId:
...
+        Breaker
+        | Chp
+        | Converter
+        | CryptoMiner
+        | Electrolyzer
+        | Hvac
+        | Meter
+        | Precharger
+        | SteamBoiler
+        | WindTurbine
```
shsms, line 69: Maybe this can be an alias

src/frequenz/client/common/microgrid/electrical_components/proto/v1alpha8/_electrical_component.py
```diff
@@ -111,7 +143,8 @@ class _ElectricalComponentBaseData(NamedTuple):
     lifetime: Lifetime
     metric_config_bounds: dict[Metric | int, Bounds]
     category_specific_info: dict[str, Any]
-    operational_mode: ElectricalComponentOperationalMode | int
+    provides_telemetry: bool | None
+    accepts_control: bool | None
```
shsms, line 147: this is a nice idea, but it also affects how this info reaches the rust component graph in a generic way, because it has to be fed by both rust and python.

Not a major inconvenience though, just need a translation layer back to enum in the python bindings.

shsms (approved):

tests/microgrid/electrical_components/test_simple_components.py
llucax: It could be, but it is used only here, so I'm not sure if it is worth it.

src/frequenz/client/common/microgrid/electrical_components/proto/v1alpha8/_electrical_component.py
llucax: What do you mean about "generic way", maybe you can give more details or point me to some code to understand it better?

shsms: component graph will need to get this from the rust client and from the python client.  If they both have the enum, then it is easy to give it to the cg.  If one of them is diverging, then we have to translate it back.

No code yet, hope to start soon, though.

shsms: I mean we're already doing this kind of dance translating from python category class to rust enum, so doing similar stuff for operational mode also doesn't sound too bad.

https://github.com/frequenz-floss/frequenz-microgrid-component-graph-python/blob/126e80fa3c22f6f4fab643c59fbda60b6a460520/src/category.rs#L238-L240

llucax:
OK, submitted the 3rd leg of the trilogy.

* #222

llucax:
I still have a small TODO list of things to polish, this trilogy is basically the last work I plan for v0.4.1.

<details>
<summary>TODO</summary>

- [ ] Check differences with assets API wrappers and apply any relevant changes (important, maybe we need to import `raise_on_error` stuff before the v0.4.1 to be able to use this in the assets client)
- [ ] Review minor/major issues, remove spurious minor issues (probably will need to go to v0.5.0 if we want to be strict about breaking changes).
- [ ] Update `metric_config_bounds` from protobuf to store UNSPECIFIED as an int (0) instead of dropping it.
- [ ] For `UnspecifiedValueError` and `UnrecognizedValueError`, add a field for the affected field name, and the target enum type, and include both in the error message. Rename both to include `Enum`: `UnspecifiedEnumValueError`, `UnrecognizedEnumValueError`.
- [ ] Extend accessors created in phase A to also raise a `UnrecognizedEnumValueError` where appropriate. For boolean attributes this means also changing the type from `bool | None` to `bool | int`, so we can distinguish between recognized values (`True`/`False`), `0` (`UNSPECIFIED`), and other unrecognized values (other int values).
- [ ] Look for "Attributes:" in class docstrings and change it to use the mkdocstrings style to add a docstring to attributes instead. Even if it is not very "standard Python" it is still valid python, it is supported much better by mkdocstrings.
- [ ] v0.5.0: Change the default for `TestEventParity.absent_members` to be frozenset({"UNSPECIFIED"}), remove `deprecated_members`.
- [ ] Add tests for `src/frequenz/client/common/test/enum_parity.py`
- [ ] Update `energy_market_code_type_from_proto()` and other enum conversion functions that have `UNSPECIFIED` deprecated to return `0` instead of `UNSPECIFIED` for the UNSPECIFIED case, and update the tests accordingly.
- [ ] Inherit unrecognized/unspecified battery/inverter/ev charger electrical component classes also from ProblematicElectricalComponent
- [ ] LOW PRIO: The _allow_construction gate: it works, but adds a second construction check alongside the existing __new__ abstractness guard, with subclasses needing to remember super().__post_init__() (e.g. GridConnectionPoint) or they silently skip the gate. Research if there is a way to merge both gates in `__new__` or `__init_subclass__` to avoid subclasses needing to remember to call super().__post_init__() and silently skipping the gate.

</details>

Then I plan to backport deprecations to v0.3.9 and release v0.5.0 removing all the deprecated. Then I will try to upgrade the clients directly from v0.3.9 to v0.5.0 if possible.

I will merge this PR now as it looks like there is enough consensus, and also so you can have a look at the next PRs, I know when they include commits from other PRs they are hard to review.

PR #222, opened 2026-06-25
Deprecate `ElectricalComponentCategory` and `XxxType` enums

### LINKED ISSUE #223: Remove all `UNSPECIFIED` enum values and low-level enums

llucax:
This PR removes the electrical component category and type enums and leaves the `ElectricalComponent` class hierarchy as the canonical representation of electrical components and single source of truth for electrical component types.

It basically:

* Deprecates the `ElectricalComponentCategory` and `ElectricalComponentType` wrapper enums.
* Deprecates the `BatteryType`, `EvChargerType` and `InverterType` wrapper enums.
* Deprecates the `category` of the `ElectricalComponent` class.
* Deprecates the `type` for `Battery`, `EvCharger` and `Inverter` classes.
* Adds `electrical_component_class_from_proto()` and `electrical_component_class_to_proto()` functions to convert between the `ElectricalComponent` class hierarchy and the protobuf enum representation of electrical component categiries and types so client implementers have an easy way to replace the deprecated wrapper enums with the new class hierarchy.

It also adds some missing electrical component classes so they are not wrapped as unrecognized electrical components.

Part of #223.


llucax:
Converting to draft as it is based on #221.

llucax:
The `electrical_component_class_from/to_proto()` are pretty convoluted. Opus choked a bit on it, and even when it did most of the structure, I needed to do the real implementation myself to get the type-hinting right (there is only one gap I could not solve).

But well, is a one time thing, adding more components should be easy even if the code is scary.

ela-kotulska-frequenz (commented):
LGTM , however I am not familiar with this code at all.

llucax:
Ready for review.

llucax:
Maybe it is a good idea to look at [the release notes changes](https://github.com/frequenz-floss/frequenz-client-common-python/pull/222/changes#diff-68d24fa81558aae3d8c59e2aa57a4fa719ea3b04d7fa14beff45f16f00858f50) first to have a better idea of where this is going.

Marenz (approved):
Clean deprecation with thorough tests, proper warning hygiene, and a clear migration path. The class-hierarchy approach is a better fit for the domain than enum switches. The 11 `@overload` signatures are verbose but necessary for mypy — worth the type safety.

PR #225, opened 2026-06-26
Deprecate remaining `UNSPECIFIED` enum members

### LINKED ISSUE #223: Remove all `UNSPECIFIED` enum values and low-level enums

llucax:
Deprecate `ElectricalComponentDiagnosticCode.UNSPECIFIED`, `ElectricalComponentStateCode.UNSPECIFIED`, and `Event.UNSPECIFIED`.

These enums didn't have any conversion functions yet, nor are used by any other code in this repository, so it is just a warning for downstream projects to adapt to not having the `UNSPECIFIED` values in the future, and use the `int` `0` value instead if necessary.

Part of #223.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -43,6 +43,9 @@
     * `frequenz.client.common.grid.EnergyMarketCodeType`
     * `frequenz.client.common.metrics.Metric`
     * `frequenz.client.common.metrics.MetricConnectionCategory`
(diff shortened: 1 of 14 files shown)


llucax:
Blocked because it is based on top of #222.

llucax:
Ready for review.

shsms (approved):

PR #232, opened 2026-07-02
Replace `Bounds` and `Lifetime` with `frequenz.core.math.Interval`

llucax:
Wrapping `Bounds` and `Lifetime` is bringing protobuf low-level concepts into the high-level API.

These 2 messages provide an interval with a start and an end, exactly what `frequenz.interval.Interval` provides.

We replace both:

- `Bounds` via a deprecation process, as it was already released.
- `Lifetime` via a direct replacement, as it was not yet released.

The deprecated `Bounds` class now inherits from `Interval`, so it can be passed in places where an `Interval` is expected. New `bounds_from_proto*2()` functions are provided to convert from protobuf messages to `Interval` objects, and the old `bounds_from_proto*()` functions (without the `2` suffix) are deprecated.

The `Lifetime` type is completely removed, the `lifetime_from_proto*()` functions now return an `Interval` instead.


llucax:
It will be a little simpler after https://github.com/frequenz-floss/frequenz-core-python/pull/178 is merged and released.

Marenz (approved):

llucax:
We had an internal discussion and @shsms was concerned about removing the domain-specific language from `Bounds` (`lower`/`upper` vs `start`/`end`). To move things forward I will close this one and keep things as they are, and add a `BoundsSet` in the future that is basically a copy of the [`IntervalSet` proposed to `core`](https://github.com/frequenz-floss/frequenz-core-python/pull/179) but specialized for bounds, and in the future we see if we abstract it in some way.

PR #233, opened 2026-07-03
Remove unreleased electrical component category/type wrappers

llucax:
The earlier deprecation PR #222 deprecated `ElectricalComponentCategory` and the `XxxType` enums and `.category` and `.type` attributes, but most of those symbols were yet unreleased, so deprecating them makes no sense.

This PR remove the unreleased deprecated symbols:

- `ElectricalComponent.category` and `(Battery|Inverter|EvCharger).type` attributes, except for `Unrecognized*` classes and `MismatchesCategoryElectricalComponent`, as they carry the raw protobuf value that could be used for forward compatibility.
- `BatteryType`, `InverterType`, `EvChargerType` enums, and their protobuf conversion functions.

`ElectricalComponentCategory` and its proto converters are kept as deprecated because they were already released and still present in v0.4.0.

Part of #223.


shsms:
will wait for the rebase before reviewing.

llucax:
Rebased.

llucax:
I was very silly not to realize in #222 that most of the stuff I deprecated was actually never released (in this repo).

ela-kotulska-frequenz (approved):

PR #237, opened 2026-07-06
Raise `UnrecognizedEnumValueError` from boolean accessors

llucax:
The boolean accessors `Microgrid.is_active()`, `ElectricalComponent.provides_telemetry()`, and `accepts_control()` wrap the `MicrogridStatus` and `ElectricalComponentOperationalMode` proto enums into a clean `bool`.

So far they only raised `UnspecifiedEnumValueError`, conflating the `UNSPECIFIED` proto sentinel with any other unknown protobuf enum value the server might send.

That made forward-compatibility silent: a future server-side mode would either disappear or look identical to `UNSPECIFIED` at the call site.

Align them with other enum values accessors (`get_*()`): the underlying low-level fields now hold a `bool` for known states, the raw `int` `0` for `UNSPECIFIED`, and the raw `int` value unchanged for any unrecognized server-side value.

### DIFF
--- a/src/frequenz/client/common/microgrid/electrical_components/proto/v1alpha8/_electrical_component.py
+++ b/src/frequenz/client/common/microgrid/electrical_components/proto/v1alpha8/_electrical_component.py
@@ -797,7 +797,7 @@ def electrical_component_class_from_proto(
 # Message converters (full protobuf message ↔ instance)
 # ============================================================================
 
-_BOOLS_BY_OPERATIONAL_MODE: dict[int, tuple[bool | None, bool | None]] = {
+_BOOLS_BY_OPERATIONAL_MODE: dict[int, tuple[bool, bool]] = {
     electrical_components_pb2.ELECTRICAL_COMPONENT_OPERATIONAL_MODE_INACTIVE: (
         False,
         False,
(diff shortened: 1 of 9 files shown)


src/frequenz/client/common/microgrid/electrical_components/proto/v1alpha8/_electrical_component.py
```diff
@@ -817,18 +817,22 @@ def electrical_component_class_from_proto(
 }
 
 
-def _operational_mode_to_bools(value: int) -> tuple[bool | None, bool | None]:
+def _operational_mode_to_bools(value: int) -> tuple[bool | int, bool | int]:
```
daniel-zullo-frequenz: The return type seems wrong or too generic to me because I think the return type will be either a tuple[bool,bool] or tuple[int,int] but not a combination of bool and int within the tuple (e.g. tuple[bool,int]

daniel-zullo-frequenz (commented):
A minor comment to check for, LGTM otherwise

src/frequenz/client/common/microgrid/electrical_components/proto/v1alpha8/_electrical_component.py
daniel-zullo-frequenz:
```suggestion
def _operational_mode_to_bools(value: int) -> tuple[bool, bool] | tuple[int, int]:
```

llucax: True, it is harmless anyway because it is internal and at the end we handle them independently, but I will update to `tuple[bool, bool] | tuple[int, int]`.

llucax:
Updated.

daniel-zullo-frequenz (approved):

PR #241, opened 2026-07-06
Improve `DeliveryArea` validation and error handling

### LINKED ISSUE #251: Standardize `<invalid:VALUE>` across `__str__`
### What's needed?

After #241, `InvalidDeliveryArea.__str__` uses an explicit `<invalid:VALUE>` marker to flag invariant-violating fields (empty `code`, `code_type=0`). The same convention is being adopted in #248 (`InvalidLatitude/Longitude/CountryCode`, tracked via review comment [r3589470688](https://github.com/frequenz-floss/frequenz-client-common-python/pull/248#discussion_r3589470688)).

Several other wrapper types across the library carry malformed / unspecified / unrecognized data but don't yet follow this convention. They should be updated so `grep '<invalid:'` on logs finds every invariant violation regardless of which type produced it.

Rationale for the `<invalid:VALUE>` marker (over `!` prefix/suffix, `❌`, class-name-only, or bare `repr()`): ASCII + greppable, composes per field, no collision with legitimate content. Full write-up at [r3589470688](https://github.com/frequenz-floss/frequenz-client-common-python/pull/248#discussion_r3589470688) and reference implementation at [commit `5688575`](https://github.com/frequenz-floss/frequenz-client-common-python/pull/241/commits/5688575e22a449a8fb0fe8937fc0a0cac0de14f0).

### Proposed solution

**Existing wrapper types in this repo (no active PR):**

| Class | File | Current | Needs |
|---|---|---|---|
| `MetricConnection.__str__` | `src/frequenz/client/common/metrics/_sample.py` | Renders raw `int` bare for unspecified/unknown `category`, though `get_category()` raises on `0 \| MetricConnectionCategory.UNSPECIFIED`. The existing wrapping on the known case uses `<CATEGORY=NAME>` (uppercase `=`), inconsistent with the new `<invalid:VALUE>` (lowercase `:`). | Render `<invalid:0>` for `0 \| MetricConnectionCategory.UNSPECIFIED` to match `get_category()`; normalize (or drop) the `<CATEGORY=NAME>` wrapping on the known case. |
| `UnrecognizedElectricalComponent` | `src/frequenz/client/common/microgrid/electrical_components/_problematic.py` | Carries `category: int` from the wire; not visible in the inherited `ElectricalComponent.__str__` (only the class name signals the problem). | Override `__str__` to expose the raw value as `<invalid:{category}>`. |
| `MismatchedCategoryElectricalComponent` | `src/frequenz/client/common/microgrid/electrical_components/_problematic.py` | Same. | Same. |
| `UnrecognizedBattery` | `src/frequenz/client/common/microgrid/electrical_components/_battery.py` | Carries `type: int`; not visible in inherited `__str__`. | Override `__str__` to expose the raw value as `<invalid:{type}>`. |
| `UnrecognizedEvCharger` | `src/frequenz/client/common/microgrid/electrical_components/_ev_charger.py` | Same. | Same. |
| `UnrecognizedInverter` | `src/frequenz/client/common/microgrid/electrical_components/_inverter.py` | Same. | Same. |

**In-flight PR needing alignment:**

- **`InvalidLifetime`** (#249 `err-lifetime` branch) — needs a design decision. Unlike the per-field violations above, `InvalidLifetime`'s invariant is *relational* (`start > end`), not per-field. Suggested: add a `__str__` on `BaseLifetime` rendering `[{start_time}, {end_time}]`, override on `InvalidLifetime` to wrap the whole rendering: `<invalid:[{start_time}, {end_time}]>`.

Tests for the classes above need updating to match the new `__str__` outputs.

### Use cases

- **Log triage**: `grep '<invalid:'` across logs finds every invariant violation regardless of type. Currently readers need to know each wrapper's convention (or that a bare `0` in `MetricConnection[0]` means "unspecified" rather than "category-0").
- **Consistency for downstream users**: application code that formats or logs any of these types sees the same shape of invalid marker regardless of what produced it.

### Alternatives and workarounds

Considered `!` prefix/suffix, `❌` emoji, class-name-only marker, and bare `repr()` fallback. Full comparison and rationale: [PR #248 r3589470688](https://github.com/frequenz-floss/frequenz-client-common-python/pull/248#discussion_r3589470688).

### Additional context

- Reference implementation: [`5688575` on `name-delivery-area`](https://github.com/frequenz-floss/frequenz-client-common-python/pull/241/) — `InvalidDeliveryArea.__str__` overrides with per-field `<invalid:VALUE>` markers; `DeliveryArea` keeps a simple `__str__` (invariants trusted on the valid class).
- Related PRs: #241 (grid, landed), #248 (types/location, tracked via review comment).

llucax:
Adds a new delivery-area class hierarchy:

* `frequenz.client.common.grid.BaseDeliveryArea` — abstract common supertype of the two concrete leaves; not directly instantiable.
* `frequenz.client.common.grid.DeliveryArea` — well-formed delivery area (retroactively made a subclass of `BaseDeliveryArea`.
* `frequenz.client.common.grid.InvalidDeliveryArea` — malformed wire data; same fields as `DeliveryArea` with no invariants enforced, so callers can inspect whatever the server actually sent.

Also adds `frequenz.client.common.grid.proto.v1alpha8.delivery_area_from_proto2` returning `DeliveryArea | InvalidDeliveryArea`. This is the replacement for `delivery_area_from_proto` (which is deprecated now) and is what `microgrid_from_proto` uses internally.

This deprecates construction of `frequenz.client.common.grid.DeliveryArea` with invalid data. A well-formed `DeliveryArea` has a non-empty `code` and a non-`UNSPECIFIED` `code_type`. Constructing one with invalid data currently emits a `DeprecationWarning`; a future release will replace the warning with a hard `ValueError`. Prefer `delivery_area_from_proto2` to load delivery areas from the wire — malformed messages become `InvalidDeliveryArea` instances instead.

Part of #239 and #251.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -44,6 +44,14 @@
 
     Users are encouraged to switch from direct field access to the new `get_*()` methods (see New Features), which provide a safer way to handle unspecified or unrecognized values.
 
+* `frequenz.client.common.grid.proto.v1alpha8.delivery_area_from_proto` is now deprecated; use `delivery_area_from_proto2` instead.
+
+    The new converter returns `DeliveryArea | InvalidDeliveryArea` and surfaces malformed wire data at the type level rather than silently constructing a `DeliveryArea` with invalid content. The old converter continues to work but emits a `DeprecationWarning`.
+
+* `frequenz.client.common.grid.DeliveryArea` construction with invalid data is deprecated. Please construct only valid `DeliveryArea` objects.
+
+    A well-formed `DeliveryArea` has a non-empty `code` (a non-`UNSPECIFIED` `code_type` is currently accepted as valid for practical reasons but will be considered invalid in the future). Constructing one with invalid data currently emits a `DeprecationWarning`; a future release will replace the warning with a hard `ValueError`. Prefer `delivery_area_from_proto2` to load delivery areas from the wire — malformed messages become `InvalidDeliveryArea` instances instead.
+
 ## New Features
 
 * Added 4 new electrical component classes for categories that previously collapsed into `UnrecognizedElectricalComponent`:
@@ -68,11 +76,19 @@
     * `frequenz.client.common.UnspecifiedEnumValueError` for unspecified enum values (raw `0` or the deprecated member).
     * `frequenz.client.common.UnrecognizedEnumValueError` for enum members not yet recognized by the library. Carries the raw integer value in its `value` attribute.
 
-* Added safe convenience getters that raise the new exceptions for unspecified or unrecognized values:
+* Added safe convenience getters that raise the new exceptions for unspecified, unrecognized, missing or invalid values:
 
     * `frequenz.client.common.grid.DeliveryArea.get_code_type()`
     * `frequenz.client.common.metrics.MetricConnection.get_category()`
     * `frequenz.client.common.metrics.MetricSample.get_metric()`
+
+* Added new delivery-area class hierarchy:
+
+    * `frequenz.client.common.grid.BaseDeliveryArea` — abstract common supertype of the two concrete leaves; not directly instantiable.
+    * `frequenz.client.common.grid.DeliveryArea` — well-formed delivery area (retroactively made a subclass of `BaseDeliveryArea`).
+    * `frequenz.client.common.grid.InvalidDeliveryArea` — malformed wire data; same fields as `DeliveryArea` with no invariants enforced, so callers can inspect whatever the server actually sent.
+
+* Added `frequenz.client.common.grid.proto.v1alpha8.delivery_area_from_proto2` returning `DeliveryArea | InvalidDeliveryArea`. This is the replacement for the now-deprecated `delivery_area_from_proto`.
 
 * Added a new `frequenz.client.common.types.Lifetime` type together with the `frequenz.client.common.types.proto.v1alpha8.lifetime_from_proto` conversion function.
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -50,7 +50,7 @@
 
 * `frequenz.client.common.grid.DeliveryArea` construction with invalid data is deprecated. Please construct only valid `DeliveryArea` objects.
 
-    A well-formed `DeliveryArea` has a non-empty `code` (a non-`UNSPECIFIED` `code_type` is currently accepted as valid for practical reasons but will be considered invalid in the future). Constructing one with invalid data currently emits a `DeprecationWarning`; a future release will replace the warning with a hard `ValueError`. Prefer `delivery_area_from_proto2` to load delivery areas from the wire — malformed messages become `InvalidDeliveryArea` instances instead.
+    A well-formed `DeliveryArea` has a non-empty `code` and a specified (non-`UNSPECIFIED`) `code_type`. Constructing one with invalid data currently emits a `DeprecationWarning`; a future release will replace the warning with a hard `ValueError`. To opt into the upcoming behavior right now, pass `_raise_on_invalid=True` to the constructor. Prefer `delivery_area_from_proto2` to load delivery areas from the wire — malformed messages become `InvalidDeliveryArea` instances instead.
 
 ## New Features
 
--- a/src/frequenz/client/common/grid/_delivery_area.py
+++ b/src/frequenz/client/common/grid/_delivery_area.py
@@ -66,6 +66,12 @@ class BaseDeliveryArea:
 
     This code could be extended in the future, in case an unknown code type is
     encountered, a plain integer value is used to represent it.
+
+    Tip:
+        This is the lower-level accessor; when working with a valid
+        [`DeliveryArea`][...DeliveryArea], prefer
+        [`get_code_type`][...DeliveryArea.get_code_type] to obtain a known
+        member or a clear error.
     """
 
     # pylint: disable-next=unused-argument
@@ -77,7 +83,7 @@ class BaseDeliveryArea:
 
 
 @dataclass(frozen=True, kw_only=True)
-class DeliveryArea:
+class DeliveryArea(BaseDeliveryArea):
     """A geographical or administrative region where electricity deliveries occur.
 
     DeliveryArea represents the geographical or administrative region, usually defined
@@ -95,19 +101,6 @@ class DeliveryArea:
         [EIC](https://en.wikipedia.org/wiki/Energy_Identification_Code) (Energy
         Identification Code). [List of
         EICs](https://www.entsoe.eu/data/energy-identification-codes-eic/eic-approved-codes/).
-    """
-
-    code: str | None
-    """The code representing the unique identifier for the delivery area."""
-
-    code_type: EnergyMarketCodeType | int
-    """Type of code used for identifying the delivery area itself.
-
-    This code could be extended in the future, in case an unknown code type is
-    encountered, a plain integer value is used to represent it.
-
-    This is the lower-level, forward-compatible accessor; prefer
-    `DeliveryArea.get_code_type()` to obtain a known member or a clear error.
     """
 
     def __str__(self) -> str:
--- a/src/frequenz/client/common/grid/_delivery_area.py
+++ b/src/frequenz/client/common/grid/_delivery_area.py
@@ -94,6 +94,14 @@ class DeliveryArea(BaseDeliveryArea):
     location. Delivery areas can have different codes based on the jurisdiction in
     which they operate.
 
+    Warning: Construction of invalid instances is deprecated
+        A well-formed `DeliveryArea` carries a non-empty [`code`][.code].
+        Constructing one with data that violates this invariant is
+        **deprecated**, and will raise a [`ValueError`][] in a future release.
+
+        In the future, delivery areas with an unspecified [`code_type`][.code_type]
+        will also be considered invalid.
+
     Note: Jurisdictional Differences
         This is typically represented by specific codes according to local jurisdiction.
 
@@ -102,6 +110,16 @@ class DeliveryArea(BaseDeliveryArea):
         Identification Code). [List of
         EICs](https://www.entsoe.eu/data/energy-identification-codes-eic/eic-approved-codes/).
     """
+
+    def __post_init__(self) -> None:
+        """Warn if this instance carries invalid data."""
+        if not self.code:
+            warnings.warn(
+                "Constructing a DeliveryArea without a `code` is deprecated and will raise "
+                "a `ValueError` in a future release",
+                DeprecationWarning,
+                stacklevel=3,
+            )
 
     def __str__(self) -> str:
         """Return a human-readable string representation of this instance."""
--- a/src/frequenz/client/common/grid/_delivery_area.py
+++ b/src/frequenz/client/common/grid/_delivery_area.py
@@ -66,7 +66,14 @@ class BaseDeliveryArea:
     """
 
     code: str | None
-    """The code representing the unique identifier for the delivery area."""
+    """The code representing the unique identifier for the delivery area.
+
+    Warning: Using `None` is deprecated
+        This field is required for a well-formed `DeliveryArea`, so we are
+        making this more explicit by deprecating the use of `None` here. In the
+        future, `| None` will be removed so passing `None` will fail type
+        checking.
+    """
 
     code_type: EnergyMarketCodeType | int
     """Type of code used for identifying the delivery area itself.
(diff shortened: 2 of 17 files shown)


llucax:
Ready for review again. @cwasicki Please confirm this makes sense to you.

I was wondering if we should bake the "default" delivery area code directly here if it is unspecified, so we can already make it logically as if it was required, instead of having downstream users having to care about this transitional phase where a missing delivery area code is OK (and some default should be assumed when this happens).

What do you think?

src/frequenz/client/common/grid/_delivery_area.py
```diff
@@ -126,3 +190,65 @@ def get_code_type(self) -> EnergyMarketCodeType:
...
+    Represents delivery area data that fails the invariants required for a
+    well-formed [`DeliveryArea`][..DeliveryArea]. Callers can inspect the raw
+    fields to recover partial information.
+
+    This class does not enforce any invariants on construction.
+    """
+
+    def __str__(self) -> str:
+        """Return a human-readable string representation of this instance."""
+        code = self.code or "❌"
```
cwasicki: Is this emoji really intended? I would avoid using emojis in our core libraries. In fact I would even avoid any non-ASCII characters. Personally, I wouldn't even know how to grep for this in logs, nor how to code this without copy-paste or AI.

cwasicki:
> I was wondering if we should bake the "default" delivery area code directly here if it is unspecified, 

Do you mean the European one?

src/frequenz/client/common/grid/_delivery_area.py
```diff
@@ -126,3 +190,65 @@ def get_code_type(self) -> EnergyMarketCodeType:
...
+        match self.code_type:
+            case EnergyMarketCodeType():
+                code_type = self.code_type.name
+            case 0:
+                code_type = "type=❌"
+            case int() as code_type:
+                code_type = f"type={code_type}"
+        return f"{code}[{code_type}]"
```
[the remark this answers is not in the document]
llucax: Fixed.

src/frequenz/client/common/grid/proto/v1alpha8/_delivery_area.py
```diff
@@ -75,4 +91,49 @@ def delivery_area_from_proto(message: delivery_area_pb2.DeliveryArea) -> Deliver
             message,
         )
 
-    return DeliveryArea(code=code, code_type=code_type)
+    # `DeliveryArea` emits a `DeprecationWarning` when constructed with
+    # invalid data. This function is scheduled to be marked `@deprecated`
+    # itself, at which point callers will see the outer notice pointing
+    # to `delivery_area_from_proto2`. Suppress the inner warning here so
+    # we don't double-warn.
+    with warnings.catch_warnings():
```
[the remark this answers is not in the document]
llucax: Fixed.

src/frequenz/client/common/microgrid/_microgrid.py
```diff
@@ -40,8 +48,17 @@ class Microgrid:  # pylint: disable=too-many-instance-attributes
...
+    delivery_area: DeliveryArea | InvalidDeliveryArea | None
+    """The delivery area where the microgrid is located.
+
+    `None` means the field was not set on the wire. An
+    [`InvalidDeliveryArea`][....grid.InvalidDeliveryArea] means the wire
+    carried a delivery area that fails its invariants.
+
+    Tip:
+        This is the lower-level field; prefer [`get_delivery_area()`][..get_delivery_area]
+        to obtain a valid [`DeliveryArea`][....grid.DeliveryArea] or a clear error.
```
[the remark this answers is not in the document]
llucax: No, docs tests are passing, this is bogus.

RELEASE_NOTES.md
```diff
@@ -68,12 +76,20 @@
     * `frequenz.client.common.UnspecifiedEnumValueError` for unspecified enum values (raw `0` or the deprecated member).
     * `frequenz.client.common.UnrecognizedEnumValueError` for enum members not yet recognized by the library. Carries the raw integer value in its `value` attribute.
 
-* Added safe convenience getters that raise the new exceptions for unspecified or unrecognized values:
+* Added safe convenience getters that raise the new exceptions for unspecified, unrecognized, missing or invalid values:
 
     * `frequenz.client.common.grid.DeliveryArea.get_code_type()`
     * `frequenz.client.common.metrics.MetricConnection.get_category()`
     * `frequenz.client.common.metrics.MetricSample.get_metric()`
```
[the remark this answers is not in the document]
llucax: Yeah, the whole class is new, we don't need to document every method of a new class here.

llucax:
> Do you mean the European one?

I guess so, whatever you are using if you get no delivery area code.

src/frequenz/client/common/grid/_delivery_area.py
llucax: Believe it or not, this was added by me, not AI :laughing: 

I thought it would be very useful to have a clear visual indication when data is not valid, but I see your point about grepping being more complicated.

What do you suggest to use instead? Just a normal "X"? "!"? I think AI originally used "?" and I didn't even realized there was a "?" there when I decided to add the emoji. The tests were failing because I didn't see (or removed) the "?" from the tests.

cwasicki: Can't it be `"None"`?

cwasicki:
> I guess so, whatever you are using if you get no delivery area code.

For the foreseeable future this would indeed be convenient. But not sure if we should really silently default to it.

llucax:
> For the foreseeable future this would indeed be convenient. But not sure if we should really silently default to it.

I feel the same, but if we stay on the practical side, actually having this behaviour in a common library means updating should be much easier when we make sure not having a delivery code type is really an error, as we only need to update one place instead of tracking all places where this ad-hoc default was added.

But I guess you are the main user of this, so I leave the decision to you.

src/frequenz/client/common/grid/_delivery_area.py
llucax: Yeah, it could be `None `, although if we are going that route maybe `""` makes more sense since it is a `str`, it should be equally greppable and we don ´t need to special case it, unless we want to keep the emoji.

How do you feel about the emoji if we keep something greappable? Like `""❌[type=0❌]` vs `""[type=0]` or  `None❌[type=0❌]` vs `None[type=0]`?

cwasicki: Tbh I would argue completely against the use of emojis, for various reasons. It's very inconvenient to code/type/grep with them. Also we cannot ensure that they are rendered at all, and if so, that they are rendered in the same way. Then it's not obvious what the emoji actually means: missing, invalid, ...

But to me the most important point is that this is a base library and you are imposing some very visually intrusive emoji into my logs. Maybe my application doesn't even care about this problem, and this just draws my attention from more severe problems. IMO these kind of formatting decisions should be made on app side and should be subject to the people who maintain it.

llucax: > Tbh I would argue completely against the use of emojis, for various reasons. It's very inconvenient to code/type/grep with them. Also we cannot ensure that they are rendered at all, and if so, that they are rendered in the same way. Then it's not obvious what the emoji actually means: missing, invalid, ...

The "might not be rendered properly" is valid, but not a huge problem if we don't **rely** on them being visible, but just used them as **visual aid**. I agree it was a mistake to try to completely replace an empty value with an emoji.

The rest sounds a little bit weird to me. It is convention, I think a red cross is a pretty clear indication something is not right, like a :heavy_check_mark: is pretty clear whatever is happening is going well.

> But to me the most important point is that this is a base library and you are imposing some very visually intrusive emoji into my logs. Maybe my application doesn't even care about this problem, and this just draws my attention from more severe problems. IMO these kind of formatting decisions should be made on app side and should be subject to the people who maintain it.

Any `str` we use will have the same issue here, it happens to all objects in all libraries. I understand you feel strongly against emojis so it feels even more intrusive/harsh, but it is not something exclusive to emojis. If you don't like it, you can always write your own string representation accessing the attributes yourself.

All of this said, I don't feel that strong to fight it, so I'm willing to use whatever you find most useful as a **default**  `str` representation.

If you find it OK, I would then avoid as much *post-processing* as possible and use something that can be put in the base class then, like:

```py
    def __str__(self) -> str:
        """Return a human-readable string representation of this instance."""
        match self.code_type:
            case EnergyMarketCodeType() as enum_code:
                code_type = enum_code.name
            case int() as int_code:
                code_type = f"type={int_code}"
            case unexpected:
                assert_never(unexpected)
        return f"{self.code or repr(self.code)}[{code_type}]"
```

Which should return  `""[type=0]` when the code is empty (we can easily use `self.code or None` here if you prefer `None[type=0]` but for me is a bit weird to use `None` here because `code` can't really be `None`, so it might be confusing, and because if there is somebody passes `code="None"` then it looks like it is wrong/empty when it is really a literal `"None"`.

cwasicki: The proposal looks good to me

src/frequenz/client/common/grid/_delivery_area.py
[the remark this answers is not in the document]
llucax: Fixed in the upcoming changes.

src/frequenz/client/common/grid/proto/v1alpha8/_delivery_area.py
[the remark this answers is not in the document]
llucax: Fixed

llucax:
Update with:
* Emojis removed
* New "standard" for invalid data in `__str__` (see #251).
* Replacing a `UNSPECIFIED` `code_type` with `EUROPE_EIC` in the new `delivery_area_from_proto2()` (overrideable via the keyword argument `replace_unspecified_code_type_with`, so if anyone needs to get the proper `UNSPECIFIED` they can, and in the future will allow different clients to move to the new required `code_type` at their own pace).

@cwasicki can you have a look and approve if it looks good to you?

src/frequenz/client/common/grid/_delivery_area.py
```diff
@@ -126,3 +211,71 @@ def get_code_type(self) -> EnergyMarketCodeType:
...
+            warnings.filterwarnings("ignore", category=DeprecationWarning)
+            match self.code_type:
+                case 0 | EnergyMarketCodeType.UNSPECIFIED:
+                    code_type = "type=<invalid:0>"
+                case EnergyMarketCodeType() as enum_code:
+                    code_type = enum_code.name
+                case int() as int_code:
+                    code_type = f"type={int_code}"
+                case unexpected:
+                    assert_never(unexpected)
```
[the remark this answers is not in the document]
llucax: We are relying on type-checking passing in all our code, so I will keep this as is.

src/frequenz/client/common/grid/proto/v1alpha8/_delivery_area.py
```diff
@@ -75,4 +91,58 @@ def delivery_area_from_proto(message: delivery_area_pb2.DeliveryArea) -> Deliver
...
+    # warning here so we don't double-warn.
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", category=DeprecationWarning)
+        return DeliveryArea(code=code, code_type=code_type)
+
+
+def delivery_area_from_proto2(
+    message: delivery_area_pb2.DeliveryArea,
+    *,
+    replace_unspecified_code_type_with: EnergyMarketCodeType = EnergyMarketCodeType.EUROPE_EIC,
```
cwasicki, line 106: Users have to update their code anyway to use this new function. Maybe we use required boolean instead to enforce an explicit decision for the fallback, i.e.
```
def delivery_area_from_proto2(
    message: delivery_area_pb2.DeliveryArea,
    replace_unspecified_code_type_with_eic: bool,
)
```

cwasicki (commented):
LGTM, I would do the default to EIC differently but not a blocker.

cwasicki (approved):

src/frequenz/client/common/grid/proto/v1alpha8/_delivery_area.py
llucax: If everybody will use the european code type, is it really useful to make the argument required? If we make it required the code needs to be updated twice. Now and then again when all delivery area messages provide a proper code type and we remove the `replace_unspecified_code_type_with_eic`. IMHO it looks like it only adds more work and doesn't buy us anything.

cwasicki: >  then again when all delivery area messages provide a proper code type and we remove the replace_unspecified_code_type_with_eic

Why would you want to remove it in future? Are you planning to remove the existing argument with its default to EIC too?

PR #247, opened 2026-07-09
Inherit value errors from a new `InvalidAttributeError`

llucax:
As we continue to add more exception that are intended for reporting invalid data in attributes, it makes sense to have a parent class with the shared attributes for all of them.

This PR adds a new `InvalidAttributeError` exception and makes all existing value errors inherit from it.

Also adds a new `MissingFieldError` in preparation for upcoming PRs.

Part of #239.


Marenz (approved):
:robot: No blocking findings. Minor residual risk: the accessor tests do not explicitly assert the exceptions' `instance` and `attr_name` metadata, but this is not approval-blocking.

PR #273, opened 2026-08-20
Preserve protobuf field names in `CategorySpecificInfo.fields`

llucax:
`electrical_component_from_proto` builds `CategorySpecificInfo.fields` with `MessageToDict`, which defaults to protobuf JSON `lowerCamelCase` keys. That contradicted the field's documented contract — "the keys are the protobuf field names" — for every multiword field: a grid connection point's `rated_fuse_current` was exposed as `ratedFuseCurrent`, so a forward-compatible caller reading the documented `fields["rated_fuse_current"]` off an unrecognized or mismatched component got a `KeyError`.

Pass `preserving_proto_field_name=True` so the keys are the `snake_case` `.proto` field names, matching both the docstring and the typed attributes (e.g. `GridConnectionPoint.rated_fuse_current`). The recognized grid connection point strips its translated field from the leftover `fields`, so update that key from `"ratedFuseCurrent"` to `"rated_fuse_current"` too; otherwise the newly snake_cased key would leak back into `fields`.

### DIFF
--- a/src/frequenz/client/common/microgrid/electrical_components/_category_specific_info.py
+++ b/src/frequenz/client/common/microgrid/electrical_components/_category_specific_info.py
@@ -34,5 +34,7 @@ class CategorySpecificInfo:
     )
     """The leftover fields not translated into typed attributes.
 
-    The keys are the protobuf field names and the values their decoded content.
+    The keys are the protobuf field names — the ``snake_case`` spelling from
+    the ``.proto`` definition (e.g. ``"rated_fuse_current"``), and the values
+    their decoded content.
     """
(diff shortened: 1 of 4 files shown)


Marenz (approved):
:robot: Correct fix for a real contract violation.

I checked all five `_leftover_info` call sites: grid was the only one keyed on a multiword field, so `"type"`, `"primary"` and `"secondary"` are unaffected by the flag and the single key update is complete.

`CategorySpecificInfo` doesn't exist at v0.4.0, so this fixes an unreleased feature and `cmd:skip-release-notes` is the right call.

PR #274, opened 2026-08-20
Suppress spurious deprecation warnings in electrical component tests

llucax:
The v1alpha8 electrical-component protobuf tests build their fixtures and parametrized cases from `ElectricalComponentCategory` members (and, in the metric-bounds test, `Metric.UNSPECIFIED`). Both enums are deprecated, so every member access raises a `DeprecationWarning` via `frequenz.core.enum`. The tests genuinely need those deprecated categories -- they exercise the protobuf `category` field the wrappers translate -- so the warnings are noise, not signal. With pytest's `once::DeprecationWarning` filter the run printed 123 of them.

Capture each deprecated member inside a minimal `warnings.catch_warnings()` / `warnings.simplefilter("ignore", DeprecationWarning)` block and reuse the captured value, so only the deprecated attribute access is suppressed and nothing around it. `test_trivial` reads its members at import time through the `parametrize` decorator, so its case list moves into a module-level suppression block (`_TRIVIAL_CASES`). This follows the pattern already used in `test_enum_deprecation.py`, leaves the intentional `pytest.deprecated_call()` / `simplefilter("error", ...)` tests untouched, and takes the suite from 123 warnings to none.

### DIFF
--- a/tests/microgrid/electrical_components/proto/v1alpha8/conftest.py
+++ b/tests/microgrid/electrical_components/proto/v1alpha8/conftest.py
@@ -3,6 +3,7 @@
 
 """Fixtures and utilities for testing electrical component protobuf conversion."""
 
+import warnings
 from datetime import datetime, timezone
 
 import pytest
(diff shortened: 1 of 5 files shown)


Marenz (approved):
:robot: The suppression is precise in the right way: it wraps only the deprecated attribute access, leaves the intentional `deprecated_call()` / `simplefilter("error", ...)` tests untouched, and the module-level `_TRIVIAL_CASES` block correctly handles the import-time `parametrize` case a per-test block could not reach.

My one objection is repetition: twelve near-identical three-line blocks. A helper in `conftest.py` would collapse them without giving up any precision:

```python
@contextlib.contextmanager
def ignore_deprecation() -> Iterator[None]:
    """Ignore `DeprecationWarning`s raised while reading deprecated enum members."""
    with warnings.catch_warnings():
        warnings.simplefilter("ignore", DeprecationWarning)
        yield
```

Call sites become `with ignore_deprecation(): category = ElectricalComponentCategory.CHP`, and there is a single place to delete once the enums are gone. Approving either way — this is a readability preference, not a correctness concern.

Note that this overlaps with #273 on `test_electrical_component_simple.py` and `test_raw_storage.py`; whichever lands first will make the other need a rebase.

PR #275, opened 2026-08-20
Use `FloatInt` from `frequenz-core`

llucax:
Use the `FloatInt` alias introduced by `frequenz-core` 1.4.0 instead of maintaining a duplicate local alias. Raise the dependency floor, remove the local module and client-common export, update annotations and tests to use the canonical import, and consolidate the mentions of `FloatInt` and why it exists in the release notes.


Marenz (approved):
:robot: Good deduplication. Verified `frequenz-core` 1.4.0 is published and exports `frequenz.core.typing.FloatInt`, and that dropping `frequenz.client.common.FloatInt` breaks nobody: the `Added \`FloatInt\`` bullet was still in the unreleased notes, so the symbol never shipped.

One nit: `tests/test_float.py` now exercises a symbol this repository doesn't own. Deleting it seems better than retargeting the import — asserting that `FloatInt` covers `float` and `int` is frequenz-core's test to keep.

PR #278, opened 2026-08-31
Add `InvalidDatetime` to represent a malformed protobuf `Timestamp`

### LINKED ISSUE #276: Return `InvalidDateTime` for invalid protobuf timestamps
## Problem

`frequenz.client.common.proto.datetime_from_proto()` currently calls `datetime.fromtimestamp()`. A protobuf `Timestamp` can carry a value that Python cannot represent, causing conversion to raise instead of preserving the received data in the return type.

For example:

```python
from google.protobuf.timestamp_pb2 import Timestamp

from frequenz.client.common.proto import datetime_from_proto

datetime_from_proto(Timestamp(seconds=253402300800))
# ValueError: year 10000 is out of range
```

This exception also escapes higher-level converters such as `metric_sample_from_proto()`, contrary to the goal of representing malformed wire data explicitly.

## Proposed change

- Add a protobuf-independent `InvalidDateTime` wrapper that preserves the raw timestamp data needed for inspection.
- Add a `datetime_from_proto2()` if `datetime_from_proto()` was already released that returns `datetime | InvalidDateTime` instead of raising for invalid or unrepresentable protobuf timestamps. Change the current function instead if it is unreleased.
- Propagate the union through all wrappers and converters that use `datetime_from_proto/2()`, including metric samples, microgrids, and lifetimes.
- Add semantic accessors and a package-specific error where callers commonly require a valid `datetime`, following the existing invalid-field pattern.
- Handle the public API change according to the project's 0.x compatibility policy.

## Tests

Cover at least:

- valid timestamps, including timezone behavior;
- the minimum and maximum values representable by both protobuf and Python;
- values outside Python's `datetime` range;
- invalid seconds/nanoseconds combinations accepted from decoded wire data;
- higher-level converters preserving `InvalidDateTime` instead of raising;
- safe accessors returning valid values and raising typed errors for invalid ones.

## Downstream migration

Find downstream users of `datetime_from_proto()` and affected wrapper datetime fields, update their annotations and exhaustive handling, and document the migration in the release notes.

llucax:
This PR closes the last gap (hopefully) of `_from_proto()` functions that can raise on invalid protobuf data.

It adds a new wrapper type `InvalidDatetime` that can be used to represent invalid timestamps in protobuf messages, and adds a new function `datetime_from_proto2()` that will return an `InvalidDatetime` instead of raising an exception.

`MetricSample` adds a `sample_time2: datetime | InvalidDatetime` field that will be set to an `InvalidDatetime` if the protobuf message has a malformed timestamp and an accessor `get_sample_time2()` that will raise the new `InvalidDatetimeError` exception if the timestamp is invalid. The old `sample_time` field is still present as a read-only property (not a real dataclass field) but deprecated.

`Microgrid` and `InvalidLifetime` fields using `datetime` are updated to use `datetime | InvalidDatetime` without a deprecation path as they are unreleased.

The user guides are updated to include when NOT to use a base class more clearly.

Fixes #276.


RELEASE_NOTES.md
```diff
@@ -136,8 +146,15 @@
     * `frequenz.client.common.metrics.MetricConnection.get_category()`
     * `frequenz.client.common.metrics.MetricSample.get_metric()`
     * `frequenz.client.common.metrics.MetricSample.get_bounds_set()`
+    * `frequenz.client.common.metrics.MetricSample.get_sample_time()`
     * `frequenz.client.common.microgrid.electrical_components.ElectricalComponent.get_metric_config_bounds()`
```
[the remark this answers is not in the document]
llucax: `Microgrid` is unreleased.

src/frequenz/client/common/metrics/proto/v1alpha8/_sample.py
```diff
@@ -174,19 +176,32 @@ def metric_sample_from_proto_with_issues(
...
         The resulting [`MetricSample`][....MetricSample] object.
+
+    Raises:
+        ValueError: If the sample time is not a well-formed protobuf
+            `Timestamp`.
     """
-    sample_time = datetime_from_proto(message.sample_time)
+    sample_time = datetime_from_proto2(message.sample_time)
+    if isinstance(sample_time, InvalidDatetime):
+        raise ValueError(f"malformed sample_time {sample_time}")
```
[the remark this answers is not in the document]
llucax: This is subtle, for a very obscure case, and a deprecated function. Not worth fixing.

llucax:
ping @Marenz

src/frequenz/client/common/metrics/_sample.py
```diff
@@ -268,7 +292,59 @@ def __init__(
...
+                given, or if neither is given; or if both `sample_time` and
+                `sample_time2` are given, or if neither is given.
+        """
+        sample_time2 = self._resolve_sample_time(sample_time, sample_time2)
+        bounds_set = self._resolve_bounds_set(bounds_set, bounds)
+        object.__setattr__(self, "sample_time2", sample_time2)
+        object.__setattr__(self, "metric", metric)
+        object.__setattr__(self, "value", value)
+        object.__setattr__(self, "bounds_set", bounds_set)
+        object.__setattr__(self, "connection", connection)
```
daniel-zullo-frequenz: This seems correct but looks a bit weird to me, like not human-friendly and I think it could be more error prone
It can probable be set like this instead:
```suggestion
        self.sample_time2 = self._resolve_sample_time(sample_time, sample_time2)
        self.bounds_set = self._resolve_bounds_set(bounds_set, bounds)

        self.metric = metric
        self.value = value
        self.connection = connection
```

src/frequenz/client/common/microgrid/proto/v1alpha8/_microgrid.py
shsms: Commit messages saying things like below might not make sense in the long run:

```
`Microgrid` is unreleased, so the annotation widens in place; no
compatibility property is needed here.
```

daniel-zullo-frequenz (approved):
I have an optional suggestion and LGTM

src/frequenz/client/common/metrics/_sample.py
llucax: This doesn't work because it is a frozen dataclass. Maybe we should add a note about it, but it is a common practice for dataclasses and `__post_init__` mutating initial state.

daniel-zullo-frequenz: I see. I haven't noticed in the diff it is a frozen dataclass

shsms (commented):
just an optional comment.

